PR #38, opened 2015-04-28
Small translation fixes

monoman:


mariospr:
Thanks a lot for your contribution. The patch looks great to me, so let's merge it now.
Muito obrigado!

PR #39, opened 2015-05-06
Fix hot corner

rshuler:
Work around broken pressure barrier.

With the recent upgrade to X, the display server now supports
extended barriers, but the functionality is currently broken
and preventing use of the hot corner.

[endlessm/eos-shell#5006]


rshuler:
I'm closing this pull request.  The real fix to the pressure barriers is straightforward, so I'm going to go with that instead.

PR #53, opened 2015-06-19
Depend on python-lcov-cobertura and eos-node-modules

smspillaz:


ptomato:
I think this pull request picked up a lot of files that it shouldn't have.

debian/changelog
```diff
@@ -1,3 +1,9 @@
+eos-shell (3.8.0~dev1) master; urgency=low
```
ptomato: Changelog not necessary

PR #61, opened 2015-07-22
ShellAppSystem: fix a crash when removing custom desktop files

cosimoc:
When g_app_info_delete() is called, such as when removing a custom
launcher from the desktop, GIO will set the desktop ID field of the
GDesktopAppInfo to NULL automatically.
That breaks an assumption in ShellAppSystem that g_app_info_get_id()
does not change over the lifecycle of a GAppInfo. When iterating over
the app infos in response to a GIO change, we'll segfault when we hit
the deleted application.

This commit fixes the issue explicitly checking for g_app_info_get_id()
returning NULL after a change from the GAppInfoMonitor.

[endlessm/eos-shell#5379]

### DIFF
--- a/src/shell-app-system.c
+++ b/src/shell-app-system.c
@@ -112,12 +112,23 @@ static gboolean
 app_is_stale (ShellApp *app)
 {
   GDesktopAppInfo *info;
+  const char *id;
   gboolean is_stale;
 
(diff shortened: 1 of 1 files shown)


sgnn7:
@cosimoc: CR pass

PR #65, opened 2015-07-28
Makefile: Allow disabling of keybindings installation

smspillaz:
By default these keybindings install systemwide and it doesn't
make sense to change their install location. Requiring root access
to run make install is not desirable and so the user
should have been able to disable this.

[endlessm/eos-shell#5439]

### DIFF
--- a/data/Makefile.am
+++ b/data/Makefile.am
@@ -43,11 +43,14 @@ gnome-shell-theme.gresource: gnome-shell-theme.gresource.xml $(resource_files)
 resourcedir = $(pkgdatadir)
 resource_DATA = gnome-shell-theme.gresource
 
-keysdir = $(GNOME_KEYBINDINGS_KEYSDIR)
 keys_in_files =					\
 	50-gnome-shell-system.xml.in		\
 	$(NULL)
 keys_DATA = $(keys_in_files:.xml.in=.xml)
+
+if INSTALL_GNOME_KEYBINDINGS
+keysdir = $(GNOME_KEYBINDINGS_KEYSDIR)
+endif
 
 gsettings_SCHEMAS = org.gnome.shell.gschema.xml
 
--- a/configure.ac
+++ b/configure.ac
@@ -142,8 +142,15 @@ PKG_CHECK_MODULES(CALENDAR_SERVER, libecal-1.2 >= $LIBECAL_MIN_VERSION libedatas
 AC_SUBST(CALENDAR_SERVER_CFLAGS)
 AC_SUBST(CALENDAR_SERVER_LIBS)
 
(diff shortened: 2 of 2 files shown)


data/Makefile.am
```diff
@@ -43,11 +43,14 @@ gnome-shell-theme.gresource: gnome-shell-theme.gresource.xml $(resource_files)
...
 resource_DATA = gnome-shell-theme.gresource
 
-keysdir = $(GNOME_KEYBINDINGS_KEYSDIR)
 keys_in_files =					\
 	50-gnome-shell-system.xml.in		\
 	$(NULL)
+
+if INSTALL_GNOME_KEYBINDINGS
+keysdir = $(GNOME_KEYBINDINGS_KEYSDIR)
 keys_DATA = $(keys_in_files:.xml.in=.xml)
```
cosimoc: I think it'd be cleaner to define `keys_DATA` outside, since it's used unconditionally e.g. in `CLEANFILES`

data/Makefile.am
```diff
@@ -43,11 +43,14 @@ gnome-shell-theme.gresource: gnome-shell-theme.gresource.xml $(resource_files)
...
+endif INSTALL_GNOME_KEYBINDINGS
```
cosimoc: I have not seen `endif` used next to a label this way; maybe it's okay, but otherwise you could just remove it.

cosimoc:
I can see why you would need this (and actually I just tried to `make distcheck` from a regular user and hit the same problem), but I was surprised that upstream never hit a problem like this. I guess that it's common practice for gnome-shell maintainers to `make distcheck` inside JHBuild and that's why they never hit this.
How exactly would this be used?

cosimoc:
@smspillaz Any comment on the above points?

smspillaz:
@cosimoc Oops, I must have missed the mail. I'll address those comments now.

data/Makefile.am
smspillaz: Yup - if `INSTALL_GNOME_KEYBINDINGS` wasn't set it might not have been cleaned either. Good catch.

smspillaz:
@cosimoc Comments addressed, thanks for the review. PR updated.

ptomato:
Candidate for upstreaming?

smspillaz:
@ptomato Done: https://bugzilla.gnome.org/show_bug.cgi?id=753814

dbnicholson:
FWIW, the normal way to handle this is to make the directory configurable but default to the value from pkg-config. Then you use DISTCHECK_CONFIGURE_FLAGS to override it to a location under `$prefix` to satisfy the check that the user can install all files.

PR #75, opened 2015-09-11
Add link in documentation to metrics query.

Kurt-von-Laven:
The metrics query must be updated when the search provider enum is
added to.

[endlessm/eos-sdk#3609]


ptomato:
**CR pass** even though the link in the comments won't be valid until the corresponding eos-desktop PR is merged.

PR #79, opened 2015-10-05
shellDBus: don't mark remote method calls as interactive

cosimoc:
Previously, it was not possible to remove an application from the store
that was not on the desktop.
Now that we have an "installed apps" page in the app store, this
operation is possible, and the app store will call into this DBus
method to make sure the desktop icon is deleted when an application is
removed.

In such case, we should not mark the operation as "interactive",
otherwise we'll get a shell notification with an Undo button that is not
intended.

While it's not exercised by the app store at the moment, also change the
same code for folder icons.

[endlessm/eos-shell#5836]


sgnn7:
@cosimoc CR pass

PR #87, opened 2015-10-27
Shell/5890

cosimoc:
endlessm/eos-shell#5890


magcius:
Should these be upstreamed?

PR #91, opened 2015-11-02
Shell/4681

mcatanzaro:
Add support for password reset. This adds a Forgot password? button to the login screen, visible after mistyping a password and clicking the password hint button if a password hint is set, immediately after mistyping the password otherwise. The user is given a random code and prompted to contact customer support at a region-specific email address or phone number. If the login screen presents the random code 1234567, compute the corresponding unlock code with `tools/password-unlocker 1234567`.

I still need to fix the following user experience issues:

endlessm/eos-shell#5894
endlessm/eos-shell#5895
endlessm/eos-shell#5966

### DIFF
--- a/data/40-gdm.rules
+++ b/data/40-gdm.rules
@@ -1,0 +1,8 @@
+polkit.addRule(function(action, subject) {
+        if (action.id == "org.freedesktop.accounts.user-administration" &&
+            (subject.user == "gdm" || subject.user == "Debian-gdm") &&
+            subject.local &&
+            subject.active) {
+                return polkit.Result.YES;
+        }
+});
--- a/data/vendor-customer-support.ini
+++ b/data/vendor-customer-support.ini
@@ -1,0 +1,8 @@
+[Customer Support]
+Email=support@endlessm.com
+Email[es]=ayuda@endlessm.com
+Email[pt]=ajuda@endlessm.com
+Phone=+1-415-980-4041
+Phone[es_GT]=+502-4761-1055 (Claro) o +502-3243-9866 (Tigo)
+Phone[es_MX]=+52-33-4170-3675
+Phone[pt_BR]=+55 (21) 9-9430-2859
--- a/js/gdm/loginDialog.js
+++ b/js/gdm/loginDialog.js
@@ -925,7 +925,7 @@ const LoginDialog = new Lang.Class({
 
     _showPrompt: function(forSecret) {
         if (this._holdForAnswer)
-            throw new Error("Previous _showPrompt not yet finished");
+            throw new Error("Assertion failure, programmer error: previous _showPrompt not yet finished");
 
         this._sessionList.actor.hide();
         this._promptLabel.show();
--- a/js/gdm/loginDialog.js
+++ b/js/gdm/loginDialog.js
@@ -933,15 +933,20 @@ const LoginDialog = new Lang.Class({
 
         this._updateSignInButtonSensitivity(this._promptEntry.text.length > 0);
 
-        this._promptEntryTextChangedId =
-            this._promptEntry.clutter_text.connect('text-changed',
-                                                    Lang.bind(this, this._onPromptEntryTextChanged));
+        // This function can be called multiple times before we disconnect.
+        if (this._promptEntryTextChangedId == 0) {
+            this._promptEntryTextChangedId =
+                this._promptEntry.clutter_text.connect('text-changed',
+                                                        Lang.bind(this, this._onPromptEntryTextChanged));
+        }
 
-        this._promptEntryActivateId =
-            this._promptEntry.clutter_text.connect('activate', Lang.bind(this, function() {
-                this._holdForAnswer.release();
-                this._holdForAnswer = null;
-            }));
+        if (this._promptEntryActivateId == 0) {
+            this._promptEntryActivateId =
+                this._promptEntry.clutter_text.connect('activate', Lang.bind(this, function() {
+                    this._holdForAnswer.release();
+                    this._holdForAnswer = null;
+                }));
+        }
     },
 
     _updateSensitivity: function(sensitive) {
--- a/js/gdm/loginDialog.js
+++ b/js/gdm/loginDialog.js
@@ -1046,7 +1046,7 @@ const LoginDialog = new Lang.Class({
         }
     },
 
-    _onAnswerProvided: function(serviceName) {
+    _onAnswerProvided: function(verifier, serviceName) {
 
          this._updateSensitivity(false);
 
@@ -1065,9 +1065,23 @@ const LoginDialog = new Lang.Class({
              this._beginVerificationForUser(user.get_user_name());
              this._passwordResetCode = null;
          } else {
-             // FIXME: Show a message when the unlock code is incorrect. Don't reset.
-             //this._passwordResetCode = null;
-             this._reset();
+             this._updateSensitivity(true);
+             this._promptEntry.set_text('');
+             this._promptMessage.set_text(_("Your unlock code was incorrect. Please try again.").format(this._passwordResetCode));
+
+             // Use an idle so that _holdForAnswer gets cleared first.
+             let id = Mainloop.idle_add(Lang.bind(this, function() {
+                 let tasks = [function() {
+                                 return this._showPrompt('');
+                             },
+
+                             function() {
+                                 this._onAnswerProvided(verifier, serviceName);
+                             }];
+                 new Batch.ConsecutiveBatch(this, tasks).run();
+                 return GLib.SOURCE_REMOVE;
+             }));
+             GLib.Source.set_name_by_id(id, '[gnome-shell] this._onAnswerProvided');
          }
     },
 
@@ -1084,7 +1098,7 @@ const LoginDialog = new Lang.Class({
 
                      function() {
                          if (this._user)
-                             this._onAnswerProvided(serviceName);
+                             this._onAnswerProvided(verifier, serviceName);
                      }];
 
         let batch = new Batch.ConsecutiveBatch(this, tasks);
--- a/js/gdm/loginDialog.js
+++ b/js/gdm/loginDialog.js
@@ -1065,7 +1065,9 @@ const LoginDialog = new Lang.Class({
         this._promptEntry.set_text('');
         this._promptEntry.clutter_text.set_password_char('');
 
-        let tasks = [this._showPrompt,
+        let tasks = [function() {
+                         return this._showPrompt(false);
+                     },
 
                      function() {
                          let userName = this._promptEntry.get_text();
--- a/js/gdm/loginDialog.js
+++ b/js/gdm/loginDialog.js
@@ -1054,6 +1054,7 @@ const LoginDialog = new Lang.Class({
 
                      function() {
                          let userName = this._promptEntry.get_text();
+                         this._user = this._userManager.get_user(userName);
                          this._promptEntry.reactive = false;
                          return this._beginVerificationForUser(userName);
                      }];
--- a/js/gdm/loginDialog.js
+++ b/js/gdm/loginDialog.js
@@ -1111,8 +1111,7 @@ const LoginDialog = new Lang.Class({
     _handleIncorrectPasswordResetCode: function(verifier, serviceName) {
          this._updateSensitivity(true);
          this._promptEntry.set_text('');
-         this._promptMessage.set_text(
-             _("Your unlock code was incorrect. Please try again.").format(this._passwordResetCode));
+         this._promptMessage.set_text(_("Your unlock code was incorrect. Please try again."));
 
          // Use an idle so that _holdForAnswer gets cleared first.
          let id = Mainloop.idle_add(Lang.bind(this, function() {
--- a/js/gdm/loginDialog.js
+++ b/js/gdm/loginDialog.js
@@ -849,8 +849,7 @@ const LoginDialog = new Lang.Class({
     cancel: function() {
         if (this._verifyingUser)
             this._userVerifier.cancel();
-        else
-            this._reset();
+        this._reset();
     },
 
     _showPrompt: function(forSecret) {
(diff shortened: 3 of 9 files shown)


mcatanzaro:
For endlessm/eos-shell#4681

data/40-gdm.rules
```diff
@@ -0,0 +1,8 @@
+polkit.addRule(function(action, subject) {
```
cosimoc: This file could be shipped in gdm itself, and then you could substitute the actual username in it, instead of having to special-case Debian-gdm.

data/vendor-customer-support.ini
```diff
@@ -0,0 +1,8 @@
+[Customer Support]
```
cosimoc: Hmm would it be possible to use the regular translation infrastructure for this instead?

js/gdm/loginDialog.js
```diff
@@ -628,10 +633,28 @@ const LoginDialog = new Lang.Class({
...
+                                                    can_focus: true,
+                                                    child: passwordResetLabel,
+                                                    reactive: true,
+                                                    x_align: St.Align.START,
+                                                    x_fill: true });
+        this._promptBox.add(this._passwordResetButton,
+                            { x_fill: false,
+                              x_align: St.Align.START });
+        this._passwordResetButton.connect('clicked', Lang.bind(this, this._showPasswordResetPrompt));
+        this._passwordResetButton.visible = false;
```
cosimoc: You could fold this property in the constructor.

js/gdm/loginDialog.js
```diff
@@ -688,12 +711,105 @@ const LoginDialog = new Lang.Class({
...
         this._passwordHintButton.hide();
+        this._maybeShowPasswordResetButton();
+    },
+
+    _readCustomerSupportData: function() {
+        try {
+            let keyFile = new GLib.KeyFile();
+            keyFile.load_from_file(Config.PKGDATADIR + '/vendor-customer-support.ini', GLib.KeyFileFlags.NONE);
+            this._customerSupportEmail = keyFile.get_locale_string('Customer Support', 'Email', null);
+            this._customerSupportPhoneNumber = keyFile.get_locale_string('Customer Support', 'Phone', null);
```
cosimoc: Use constant strings for these

js/gdm/loginDialog.js
```diff
@@ -688,12 +711,105 @@ const LoginDialog = new Lang.Class({
...
+        } catch (e) {
+            logError(e, 'Failed to read customer support data');
+            return false;
+        }
+
+        return this._customerSupportEmail && this._customerSupportPhoneNumber;
+    },
+
+    _maybeShowPasswordResetButton: function() {
+        if (!this._customerSupportPhoneNumber || !this._customerSupportEmail) {
```
cosimoc: You could rename `_readCustomerSupportData` to something like `_ensureCustomerSupportData` and fold this check in there.

js/gdm/loginDialog.js
```diff
@@ -688,12 +711,105 @@ const LoginDialog = new Lang.Class({
...
+        Polkit.Permission.new('org.freedesktop.accounts.user-administration',
+                              Polkit.UnixProcess.new_for_owner(pid, 0, -1),
+                              null,
+                              Lang.bind(this, function (obj, result) {
+                                  try {
+                                      let permission = Polkit.Permission.new_finish(result);
+                                      if (permission.get_allowed())
+                                          this._passwordResetButton.show();
+                                  } catch(e) {
+                                      logError(e, 'Failed to determine if password reset is allowed');
```
cosimoc: Should probably hide the button here?

js/gdm/loginDialog.js
```diff
@@ -688,12 +711,105 @@ const LoginDialog = new Lang.Class({
...
+    _computeUnlockCode: function(resetCode) {
+        let checksum = new GLib.Checksum(GLib.ChecksumType.MD5);
+        checksum.update(ByteArray.fromString(resetCode));
+
+        let unlockCode = checksum.get_string();
+        unlockCode = unlockCode.replace(/\D/g, '');
+        unlockCode = unlockCode.slice(0, _RESET_CODE_LENGTH);
+
+        while (unlockCode.length < _RESET_CODE_LENGTH)
+            unlockCode += '0';
```
cosimoc: Is this something that could actually happen? I thought that MD5 checksums were always more than 7 characters.

js/gdm/loginDialog.js
```diff
@@ -725,26 +841,40 @@ const LoginDialog = new Lang.Class({
...
             this._showUserList();
+
+        if (this._holdForAnswer) {
+            this._holdForAnswer.release();
+            this._holdForAnswer = null;
+        }
     },
 
     _verificationFailed: function() {
+        // Nothing to do if we were just reset after too many failed verifications.
```
cosimoc: Don't we want to clear the other dialog attributes when `this._user` is `null`?

js/gdm/loginDialog.js
```diff
@@ -778,11 +908,13 @@ const LoginDialog = new Lang.Class({
...
         if (this._verifyingUser)
             this._userVerifier.cancel();
-        else
-            this._reset();
+        this._reset();
     },
 
     _showPrompt: function(forSecret) {
+        if (this._holdForAnswer)
+            throw new Error("Previous _showPrompt not yet finished");
```
cosimoc: Are the callers of `_showPrompt()` prepared to deal with it throwing an error now?

data/40-gdm.rules
mcatanzaro: Yes, but it wouldn't make sense to ship this file with gdm unless the logic for password resets also moves to gdm.

js/gdm/loginDialog.js
```diff
@@ -800,19 +932,29 @@ const LoginDialog = new Lang.Class({
...
     },
 
-    _prepareDialog: function(forSecret, hold) {
+    _onPromptEntryTextChanged: function() {
+        if (this._passwordResetCode == null) {
+            this._updateSignInButtonSensitivity(this._promptEntry.text.length > 0);
+        } else {
+            this._updateSignInButtonSensitivity(
+                this._promptEntry.text.length == _RESET_CODE_LENGTH &&
+                this._promptEntry.text.search(/\D/) == -1);
```
cosimoc: Not sure I fully understand why these conditions are here.

js/gdm/loginDialog.js
```diff
@@ -931,6 +1079,56 @@ const LoginDialog = new Lang.Class({
...
+         this._user = user;
+         this._beginVerificationForUser(user.get_user_name());
+         this._passwordResetCode = null;
+    },
+
+    _handleIncorrectPasswordResetCode: function(verifier, serviceName) {
+         this._updateSensitivity(true);
+         this._promptEntry.set_text('');
+         this._promptMessage.set_text(
+             _("Your unlock code was incorrect. Please try again.").format(this._passwordResetCode));
```
cosimoc: There's no place in the string to format the reset code into

js/gdm/loginDialog.js
```diff
@@ -931,6 +1079,56 @@ const LoginDialog = new Lang.Class({
...
+
+         // Use an idle so that _holdForAnswer gets cleared first.
```
cosimoc: I don't entirely follow this. Would prefer explicit clearing here if possible.

data/vendor-customer-support.ini
mcatanzaro: Yes. That would let me remove one function, but there would be some disadvantages:

* Translators become responsible for keeping the contact info correct for each locale.
* Unless you have completely separate es_GT and es_MX translations (I suspect not), then you'd need to display all the numbers for both countries. Probably fine with just three numbers, but if you add more countries in the future....
* This way, the numbers can be kept all in one place, and easily updated or reviewed for correctness, without digging into different po files.

This file could also move to a different package and be used by different components that might want to display customer support contact info.

tools/password-unlocker.js
```diff
@@ -0,0 +1,35 @@
...
+
+String.prototype.format = Format.format;
+
+if (ARGV.length != 1) {
+    print('This script should be called with a reset code as the first and only argument');
+} else if (ARGV[0].length != RESET_CODE_LENGTH) {
+    print('Invalid reset code %s; valid reset codes have length %d'.format(ARGV[0], RESET_CODE_LENGTH));
+} else if (ARGV[0].search(/\D/) != -1) {
+    print('Invalid reset code %s; code should only contain digits'.format(ARGV[0]));
+} else {
```
cosimoc: If you want to have this script for development purpose, perhaps the checksum logic should be split into an util function that is included from here?

js/gdm/loginDialog.js
mcatanzaro: You mean {visible: false}, right? OK.

js/gdm/loginDialog.js
mcatanzaro: OK

js/gdm/loginDialog.js
mcatanzaro: OK

js/gdm/loginDialog.js
mcatanzaro: I'm not sure about that... then I'd want to rename the function to _maybeShowOrHidePasswordResetButton!

data/vendor-customer-support.ini
cosimoc: On the final system, the file would contain translations regardless of where they live, unless I'm missing something? So other programs could still use them.
We manage translations globally in Transifex, so digging into different po files would not be a concern.

We can leave it like this for now, but I'm tagging @rshuler for his preference here, since he's the one usually managing translations.

js/gdm/loginDialog.js
cosimoc: The function is anonymous  - either way not a big deal

js/gdm/loginDialog.js
mcatanzaro: I should have left a comment above the regex, but the line with /\D/g is removing all the non-digit characters from the hash (to make it easier to read over the phone). So if somehow you get a MD5 hash that has fewer than _RESET_CODE_LENGTH digits, then we have a problem. I presume that will basically never happen, but the case should still be handled.

js/gdm/loginDialog.js
mcatanzaro: Nope, because they were already just cleared by _reset(). It might be more clear to write if (!this._verifyingUser) instead, but checking this._user seems worthwhile since that's the variable that's actually being used.

js/gdm/loginDialog.js
mcatanzaro: No, this is an assertion that should never happen. I couldn't find any equivalent of assert() in Gjs, and Philip Chimento suggested throwing an exception instead. I added it for debugging and decided not to remove it because I found it very useful. If this isn't idiomatic, it should be removed.

js/gdm/loginDialog.js
mcatanzaro: You should be able to enter any password with one or more characters, but you should only be able to enter a password unlock code with exactly `_RESET_CODE_LENGTH` characters. If `this._passwordResetCode != null` then the user requested a password reset and needs to enter the unlock code.

js/gdm/loginDialog.js
mcatanzaro: Yuck, good catch! The string is right; I'll remove the bogus call to `format`.

tools/password-unlocker.js
mcatanzaro: Maybe.... The value of duplicating the logic here is that it allows this script to be used if it's copied outside the eos-desktop repository, e.g. so that customer support can use it. But if we make a separate application for customer support, and never plan to use this script anywhere else, then that would make sense.

cosimoc: Good point. We can keep it like this for now.

js/gdm/loginDialog.js
ptomato: This is how I'd do it, yes — perhaps make the intention clearer with `throw new Error('Assertion failure, programmer error: previous _showPrompt not yet finished')`

js/gdm/loginDialog.js
mcatanzaro: Basically: as soon as this function returns, `this._holdForAnswer = null` will be executed, and that needs to happen before this code runs. When I wrote this code, I thought it would be less confusing to use an idle to delay execution rather than write `this._holdForAnswer = null`, which would work fine. The risk of getting rid of the idle is that it makes it possible that the assertion we discussed up above won't trigger when it should if this code is changed in the future. Also, and more importantly, it's confusing to set `this._holdForAnswer = null` any place other than right after `this._holdForAnswer.release()` is called. That's why I decided to keep the idle. But it's a tough call, and either way is fine by me.

js/gdm/loginDialog.js
mcatanzaro: OK.

js/gdm/loginDialog.js
```diff
@@ -628,10 +633,28 @@ const LoginDialog = new Lang.Class({
...
+                                                style_class: 'login-dialog-password-recovery-link' });
+        this._passwordResetButton = new St.Button({ style_class: 'login-dialog-password-recovery-button',
+                                                    button_mask: St.ButtonMask.ONE | St.ButtonMask.THREE,
+                                                    can_focus: true,
+                                                    child: passwordResetLabel,
+                                                    reactive: true,
+                                                    x_align: St.Align.START,
+                                                    x_fill: true });
+        this._promptBox.add(this._passwordResetButton,
+                            { visible: false,
```
cosimoc: I think this should go in the constructor above, not here.

mcatanzaro: Yes, clearly; thanks for the catch. I'm a bit surprised this didn't cause any problems, but I guess as part of the BoxLayout it's just ignored, and the button must be hidden by default anyway.

data/vendor-customer-support.ini
rshuler: I don't think we want the contact info in Transifex.  @mcatanzaro is correct that we do not maintain es_MX vs. es_GT translations.  And, eventually, I suspect we may need other rules for selecting the appropriate contact info.  For now, the ini file works for me as is.

PR #93, opened 2015-11-04
Move 40-gdm.rules to $(datadir)

mcatanzaro:
cosimoc:  mcatanzaro, on second thought I don't think the rules file
should go in /etc/polkit-1/rules.d
cosimoc:  should go into /usr/share/polkit-1/rules.d instead
cosimoc:  it's not a distributor override - that's what the /etc rules
are for AFAIU
[endlessm/eos-shell#4681]


mcatanzaro:
This requires another pull request for the Debian package: endlessm/eos-desktop#92

PR #101, opened 2015-12-29
issue/5928

GeorgesStavracas:
When we're running on composite mode, the icon grid,
altogether with the current shell theme, make the icons
too big to fit the available screen width.

Fix that by adding a .composite-mode style class which
reduces the icon size when running on small screen width.

[endlessm/eos-shell#5928]

### DIFF
--- a/js/ui/iconGrid.js
+++ b/js/ui/iconGrid.js
@@ -477,6 +477,11 @@ const IconGrid = new Lang.Class({
         this.saturation = new Shell.GridDesaturateEffect({ factor: 0,
                                                            enabled: false });
         this.actor.add_effect(this.saturation);
+
+        /* Setup the composite mode of the grid */
+        Main.layoutManager.connect('monitors-changed', Lang.bind(this, this._updateCompositeMode));
+
+        this._updateCompositeMode();
     },
 
     _getPreferredWidth: function (grid, forHeight, alloc) {
@@ -949,5 +954,30 @@ const IconGrid = new Lang.Class({
         }
 
         return [dropIdx, cursorLocation];
+    },
+
+    _updateCompositeMode: function() {
+        /* Given that the height of the widget varies according to the
+         * number of icons, we only check the width here.
+         */
+        this.compositeMode = Main.layoutManager.primaryMonitor &&
+                             Main.layoutManager.primaryMonitor.width < 800;
+    },
+
+    set compositeMode(mode) {
+        if (this._compositeMode != mode) {
+            this._compositeMode = mode;
+
+            /* When we're running on small screens, to make it fit 5 columns
+             * on the available space, we shall reduce the icon size. Do
+             * this by adding (or removing, in case the screen is big enough)
+             * the .composite-mode style class.
+             */
+            if (mode) {
+                this.actor.add_style_class_name('composite-mode');
+            } else {
+                this.actor.remove_style_class_name('composite-mode');
+            }
+        }
     }
 });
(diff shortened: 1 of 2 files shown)


js/ui/iconGrid.js
```diff
@@ -569,6 +569,19 @@ const IconGrid = new Lang.Class({
...
         }
+
+        /* When we're running on small screens, to make it fit 5 columns
+         * on the available space, we shall reduce the icon size. Do
+         * this by adding (or removing, case the screen is big enought)
+         * the .composite-mode style class. Given that the height of the
+         * widget varies according to the number of icons, we only check
+         * the width here.
+         */
+         if (availWidth < 800) {
```
cosimoc: Tracking the allocation here, while probably effective, sounds a bit fragile. Do you think you could instead track the size of the primary monitor as defined in `layout.js`?

js/ui/iconGrid.js
```diff
@@ -569,6 +569,19 @@ const IconGrid = new Lang.Class({
...
+             this.actor.add_style_class_name ("composite-mode");
```
cosimoc: Coding style: no space before paren and use single quotes for non-translatable strings.

js/ui/iconGrid.js
GeorgesStavracas: The problem with this approach is that IconGrid class is used in other places besides the dektop app grid. If we're going to use that code, it must be done in AllView class. Which I'm working on right now.

js/ui/iconGrid.js
```diff
@@ -477,6 +477,11 @@ const IconGrid = new Lang.Class({
         this.saturation = new Shell.GridDesaturateEffect({ factor: 0,
                                                            enabled: false });
         this.actor.add_effect(this.saturation);
+
+        Main.layoutManager.connect('monitors-changed', Lang.bind(this, function() {
+            this.compositeMode = Main.layoutManager.primaryMonitor &&
+                                 Main.layoutManager.primaryMonitor.width < 800;
```
cosimoc: You could factor this into an `updateCompositeMode()` function and also call it once after connecting to the signal. That should make it possible to avoid calling it every size allocate.

js/ui/iconGrid.js
```diff
@@ -477,6 +477,13 @@ const IconGrid = new Lang.Class({
         this.saturation = new Shell.GridDesaturateEffect({ factor: 0,
                                                            enabled: false });
         this.actor.add_effect(this.saturation);
+
+        /* Setup the composite mode of the grid */
+        Main.layoutManager.connect('monitors-changed', Lang.bind(this, function() {
```
cosimoc: You don't need to wrap `this._updateCompositeMode()` in another function here - you can just use it inside the call to `Lang.bind()`

js/ui/iconGrid.js
```diff
@@ -949,5 +956,30 @@ const IconGrid = new Lang.Class({
...
+                             Main.layoutManager.primaryMonitor.width < 800;
+    },
+
+    set compositeMode(mode) {
+        if (this._compositeMode != mode) {
+            this._compositeMode = mode;
+
+            /* When we're running on small screens, to make it fit 5 columns
+             * on the available space, we shall reduce the icon size. Do
+             * this by adding (or removing, case the screen is big enought)
```
cosimoc: `enought` -> `enough`

cosimoc: Also `case the screen` -> `in case the screen`

PR #104, opened 2016-01-13
issue/5929

GeorgesStavracas:
When the primary monitor is running on composite mode,
we should make it fit 3 rows before exposing the scroll
bar.

To fix that, add the .composite-mode style classes when
the primary monitor is too small.

[endlessm/eos-shell#5929]

### DIFF
--- a/data/theme/gnome-shell.css
+++ b/data/theme/gnome-shell.css
@@ -927,6 +927,10 @@ StScrollBar StButton#vhandle:active {
     -minimum-vpadding: 48px;
 }
 
+#searchEntry.low-resolution {
+    -minimum-vpadding: 24px;
+}
+
 #searchEntry:rtl {
     padding-left: 4px;
     padding-right: 8px;
@@ -1166,6 +1170,10 @@ StScrollBar StButton#vhandle:active {
 
 .all-apps {
     -natural-padding-top: 25px;
+}
+
+.all-apps.low-resolution {
+    -natural-padding-top: 15px;
 }
 
 .search-display > StBoxLayout,
--- a/data/theme/gnome-shell.css
+++ b/data/theme/gnome-shell.css
@@ -1123,7 +1123,7 @@ StScrollBar StButton#vhandle:active {
     -shell-grid-vertical-item-size: 118px;
 }
 
-.icon-grid.composite-mode {
+.icon-grid.low-resolution {
     -shell-grid-horizontal-item-size: 104px;
     -shell-grid-vertical-item-size: 104px;
 }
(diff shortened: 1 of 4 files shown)


data/theme/gnome-shell.css
```diff
@@ -927,6 +927,11 @@ StScrollBar StButton#vhandle:active {
     -minimum-vpadding: 48px;
 }
 
+#searchEntry.composite-mode {
+    width: 350px;
```
mariospr: Why are you increasing the width from 300px to 350px?

js/ui/viewSelector.js
```diff
@@ -65,6 +65,11 @@ const ViewsDisplayLayout = new Lang.Class({
 
         this._heightAboveEntry = 0;
         this.searchResultsTween = 0;
+        this._compositeMode = false;
```
mariospr: Although this change is motivated by the composite mode, it's actually something to be applied on low enough resolutions, regardless of whether we're using composite mode or not, so I think I would rename this variable (and similar references in the .css file) to prevent confusion :). Something like `this._lowResolutionMode` maybe?

js/ui/viewSelector.js
```diff
@@ -146,6 +151,33 @@ const ViewsDisplayLayout = new Lang.Class({
...
             this._searchResultsActor.allocate(searchResultsBox, flags);
         }
+    },
+
+    _updateCompositeMode: function() {
+        /* Given that the height of the widget varies according to the
+         * number of icons, we only check the width here.
+         */
+        this.compositeMode = Main.layoutManager.primaryMonitor &&
+                             Main.layoutManager.primaryMonitor.height < 600;
```
mariospr: Maybe move this `600` to a constant somewhere up in this file?

js/ui/viewSelector.js
```diff
@@ -146,6 +151,33 @@ const ViewsDisplayLayout = new Lang.Class({
...
+    },
+
+    set compositeMode(mode) {
+        if (this._compositeMode != mode) {
```
mariospr: I would replace this with an early return, to reduce the indentation level below. Something like this:
```javascript
    if (this._compositeMode == mode)
        return;

    [ Rest of the code ]
```

js/ui/viewSelector.js
```diff
@@ -146,6 +151,33 @@ const ViewsDisplayLayout = new Lang.Class({
...
+            this._compositeMode = mode;
+
+            /* When running on small screens, to make it fit 3 rows of icons,
+             * reduce the space above and below the search entry by adding (or
+             * removing, in case the screen is big enough) the .composite-mode
```
mariospr: Remember to update the comment if you move to use something other than "composite-mode" / "compositeMode".

mariospr:
Looks great, thanks! Merged

PR #105, opened 2016-01-18
Fix icon grid when quickly removing+adding app

joaquimrocha:
When removing an application and quickly adding it again sometimes the icon
would not be added back again because there would be no apparent changes
in the icon list. To avoid this, these changes update the icon list before
the animation and subsequent icon addition.

[endlessm/eos-shell#5849]


mariospr:
Looks good to me. I had the doubt of whether it would make more sense to do that inside the _findIconChanges() function (and then rename the funcion), but then you'd have that logic executed when calling it from iconsNeedRedraw(), so it's probably better this way.

Merged

PR #106, opened 2016-01-19
appDisplay: Fix app folder close button's position

joaquimrocha:
The app folder's close button is positioned on the top right of the
folder's boxPointer. However, when the boxPointer has its arrow on
top (when the folder pops downwards), its box will be positioned lower
and thus the close button appears a bit higher than it should be.

These changes fix that problem by placing the close button taking
into account the position of the boxPointer.

[endlessm/eos-shell#6263]

### DIFF
--- a/js/ui/appDisplay.js
+++ b/js/ui/appDisplay.js
@@ -1687,6 +1687,7 @@ const AppFolderPopup = new Lang.Class({
 
         this._boxPointer.actor.bind_property('opacity', this.closeButton, 'opacity',
                                              GObject.BindingFlags.SYNC_CREATE);
+        this._boxPointer.bin.connect('realize', Lang.bind(this, this._adjustCloseButton));
     },
 
     toggle: function() {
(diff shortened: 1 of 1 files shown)


js/ui/appDisplay.js
```diff
@@ -1696,10 +1696,20 @@ const AppFolderPopup = new Lang.Class({
             this.popup();
     },
 
+    _adjustCloseButton: function() {
+        // When the popup has its arrow on top, the close button needs to be
+        // positioned a bit lower so it's placed on the boxPointer corner's edge
+        this._boxPointer.bin.realize();
+        const marginTop = this._boxPointer.bin.get_y();
+        this.closeButton.set_y(this.closeButton.get_y() + marginTop);
```
mariospr: Would it be possible to have this logic moved into a callback for when the 'realize' signal is emitted over `this._boxPointer.bin`? I find it strange that you have to call gtk_widget_realize() explicitly here.

joaquimrocha:
I can do that.

joaquimrocha:
I've updated with and connected the function to the realize signal of the boxPointer instead.

mariospr:
Thanks, I think it makes more sense now. Merged

PR #113, opened 2016-02-08
Add a way to update an app's desktop info

joaquimrocha:
These changes add a DBus method to make the desktop aware that something about an application's desktop info has changed and, currently, uses that trigger to re-read an app's desktop info and update its icon in the desktop.
The use case is a weblink that can be added to the desktop before its favicon has been retrieved; thus, once the favicon has been downloaded and the desktop file changed, the app icon will be updated o the desktop with these changes.

[endlessm/eos-shell#4920]

### DIFF
--- a/js/ui/appDisplay.js
+++ b/js/ui/appDisplay.js
@@ -534,6 +534,8 @@ const AllView = new Lang.Class({
         this._appSystem.connect('installed-changed', Lang.bind(this, function() {
             Main.queueDeferredWork(this._allAppsWorkId);
         }));
+        this._appSystem.connect('app-info-changed', Lang.bind(this, this._onAppInfoChanged));
+
         global.settings.connect('changed::app-folder-categories', Lang.bind(this, function() {
             Main.queueDeferredWork(this._allAppsWorkId);
         }));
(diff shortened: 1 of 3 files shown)


js/ui/appDisplay.js
```diff
@@ -362,10 +362,20 @@ const EndlessApplicationView = new Lang.Class({
...
+                icon.icon.reloadIcon();
+            }
+        }
+    },
+
     addIcons: function(isHidden) {
         // Don't do anything if we don't have more up-to-date information, since
         // re-adding icons unnecessarily can cause UX problems
         if (!this.iconsNeedRedraw()) {
+            this._reloadIcons();
```
cosimoc: Just a drive-by heads up that always doing I/O from this code path unfortunately **is not** acceptable. The `iconsNeedRedraw()` system was added exactly to avoid that. It's true that reloading the visible icons only is a little better than what we were doing back then (IIRC calling `g_app_info_get_all()`) but this is still a hot path, so the smarter solution may be really what we have to do here.

ebassi: I have to agree with @cosimoc here: I/O in the compositor's main thread is pretty much going to cause interactivity issues on less powerful hardware platforms.

joaquimrocha: Okay. That makes sense. I thought it'd not be that problematic, but it definitely doesn't scale so I will go for a smarter approach.

js/ui/appDisplay.js
```diff
@@ -591,6 +593,13 @@ const AllView = new Lang.Class({
         }
      },
 
+     _onAppInfoChanged: function(app_system, app) {
+        for (let icon of this._allIcons) {
+            if (icon.app && icon.app === app)
+                icon.icon.reloadIcon();
```
ebassi: Should probably break after `reloadIcon()`; we don't allow duplicates.

joaquimrocha:
Updated after your comments @ebassi .

PR #114, opened 2016-02-09
Reimplement workspace monitor

joaquimrocha:
These changes reimplement the workspace monitor and base the logic on apps' events instead of windows's events.

[endlessm/eos-shell#4190]

### DIFF
--- a/js/ui/workspaceMonitor.js
+++ b/js/ui/workspaceMonitor.js
@@ -12,6 +12,7 @@ const WorkspaceMonitor = new Lang.Class({
     _init: function() {
         this._trackedApps = new Set();
         this._visibleApps = new Set();
+        this._startingApps = new Set();
 
         this._windowTracker = Shell.WindowTracker.get_default();
 
@@ -115,13 +116,25 @@ const WorkspaceMonitor = new Lang.Class({
         }
     },
 
+    _setAppStarting: function(app, starting) {
+        if (starting) {
+            this._startingApps.add(app);
+        } else {
+            this._startingApps.delete(app);
+        }
+    },
+
     _onAppStateChange: function(appSystem, app) {
-        if (app.get_state() == Shell.AppState.STOPPED) {
+        let app_state = app.get_state();
+
+        if (app_state == Shell.AppState.STOPPED) {
             this._setAppVisible(app, false);
             this._untrackApp(app);
         } else {
             this._trackApp(app);
         }
+
+        this._setAppStarting(app, app_state == Shell.AppState.STARTING);
     },
 
     _appHasVisibleWindows: function(app, excludeWindow) {
@@ -148,12 +161,7 @@ const WorkspaceMonitor = new Lang.Class({
     },
 
     _hasStartingApps: function() {
-        for (let app of this._trackedApps) {
-            if (app.get_state() == Shell.AppState.STARTING) {
-                return true;
-            }
-        }
-        return false;
+        return this._startingApps.size > 0;
     },
 
     get visibleApps() {
--- a/js/ui/workspaceMonitor.js
+++ b/js/ui/workspaceMonitor.js
@@ -125,16 +125,16 @@ const WorkspaceMonitor = new Lang.Class({
     },
 
     _onAppStateChange: function(appSystem, app) {
-        let app_state = app.get_state();
+        let state = app.get_state();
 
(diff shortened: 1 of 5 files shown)


js/ui/workspaceMonitor.js
```diff
@@ -10,276 +10,142 @@ const WorkspaceMonitor = new Lang.Class({
...
-        // back to the overview if the only window that was dismissed was the
-        // splash screen.
-        // See also shell_app_is_interesting_window() in src/shell-app.c, which
-        // was added for the same purpose.
-        // This code eventually should be changed to track applications instead
-        // of windows.
-        if (!Shell.WindowTracker.is_window_interesting(metaWindow) ||
-            metaWindow.get_role() == 'eos-speedwagon') {
-            return false;
+        if (app in this._trackedApps) {
```
cosimoc: Not entirely sure if this is needed here, since the same block in `_minimizeWindow` should already have run for the same window.

js/ui/workspaceMonitor.js
```diff
@@ -10,276 +10,142 @@ const WorkspaceMonitor = new Lang.Class({
...
+    _unminimizeWindow: function(shellwm, actor) {
+        let app = this._windowTracker.get_window_app(actor.meta_window);
 
-    _windowRemoved: function(metaWorkspace, metaWindow) {
-        // Remove the window from our known windows
-        // Window will not be in knownWindows if it wasn't
-        // interesting to us.
-        if (!this._windowIsKnown(metaWindow)) {
-            return;
+        if (app in this._trackedApps ) {
```
cosimoc: Extra space after `this._trackedApps`

js/ui/workspaceMonitor.js
```diff
@@ -10,276 +10,142 @@ const WorkspaceMonitor = new Lang.Class({
...
-
-        this._removeKnownWindow(metaWindow);
     },
 
-    _windowGoingInvisible: function() {
-        // Check if we're minimizing the very last window
-        if (this.visibleWindows == 1) {
-            Main.layoutManager.prepareForOverview();
+    _updateOverview: function() {
+        if (this._visibleApps.size == 0) {
```
cosimoc: (optional suggestion): I wonder if instead of always keeping track of the visible apps it wouldn't be simpler to compute it on demand from the tracked applications at the time when `_updateOverview()` is called.

src/shell-app.c
```diff
@@ -937,7 +939,10 @@ shell_app_sync_running_state (ShellApp *app)
 
   if (app->state != SHELL_APP_STATE_STARTING)
     {
-      if (app->running_state->interesting_windows == 0)
+      /* We check if the startup completed as a workaround for apps that show
+       * a window and quickly remove it before the main window shows up, which
+       * is the case with Maps. */
+      if (app->running_state->interesting_windows == 0 && app->running_state->startup_complete)
```
cosimoc: I wonder if instead of tracking the startup completion this way here you couldn't check that there are no speedwagon windows in addition to interesting windows.
I.e. (pseudo-code) ` if (app->running_state->interesting_windows == 0 && app_has_no_speedwagon_windows(app))`

joaquimrocha: I tried to avoid going through all the windows. I know there shouldn't be that much but I try to make my code optimal (yes, even if I've wrote the IO thingy yesterday :D ).

js/ui/workspaceMonitor.js
joaquimrocha: That's less optimal because we have to loop, even if there will never be too many apps...

js/ui/workspaceMonitor.js
joaquimrocha: Good catch. Since we only untrack apps when they stop and when we do, we remove then from visibleApps as well, it should not be a problem.

js/ui/workspaceMonitor.js
cosimoc: I understand, but on the other hand you would save the loops caused by the `appHasVisibleWindows()` calls. I think at the end it wouldn't make much of a difference.

src/shell-app.c
cosimoc: My suggestion to do it the other way is because it avoids relying on the fact that a speedwagon window will be there to transition to the STOPPED state. With your patch as it is, we'll only transition to STOPPED when a speedwagon window has been seen, which is a bit suboptimal.

PR #115, opened 2016-02-10
Reimplement the workspace monitor based on apps

joaquimrocha:
These changes replace the way the workspace monitor works by tracking
visible windows per application instead of per window. Besides
simplifying the code, it also fixes some issues related to launching
apps, e.g. launching Maps (marble) was sometimes opening the app with
the windows minimized.

[endlessm/eos-shell#4190]

### DIFF
--- a/js/ui/workspaceMonitor.js
+++ b/js/ui/workspaceMonitor.js
@@ -10,246 +10,25 @@ const WorkspaceMonitor = new Lang.Class({
     Name: 'WorkspaceMonitor',
 
     _init: function() {
-        this._metaScreen = global.screen;
+        this._trackedApps = new Hash.Map();
+        this._visibleApps = new Set();
 
-        this._destroyedWindows = new Hash.Map();
-        this._minimizedWindows = new Hash.Map();
-        this._knownWindows = new Hash.Map();
+        this._windowTracker = Shell.WindowTracker.get_default();
 
         this._shellwm = global.window_manager;
         this._shellwm.connect('minimize', Lang.bind(this, this._minimizeWindow));
         this._shellwm.connect('minimize-completed', Lang.bind(this, this._minimizeWindowCompleted));
         this._shellwm.connect('unminimize', Lang.bind(this, this._unminimizeWindow));
         this._shellwm.connect('map', Lang.bind(this, this._mapWindow));
-        this._shellwm.connect('destroy', Lang.bind(this, this._destroyWindow));
-        this._shellwm.connect('destroy-completed', Lang.bind(this, this._destroyWindowCompleted));
 
-        this._metaScreen.connect('workspace-switched', Lang.bind(this, this._workspaceSwitched));
+        this._metaScreen = global.screen;
         this._metaScreen.connect('in-fullscreen-changed', Lang.bind(this, this._fullscreenChanged));
 
         let primaryMonitor = Main.layoutManager.primaryMonitor;
         this._inFullscreen = primaryMonitor && primaryMonitor.inFullscreen;
-        this._visibleWindows = 0;
 
-        this._trackWorkspace(this._metaScreen.get_active_workspace(), false);
-    },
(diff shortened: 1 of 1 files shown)


js/ui/workspaceMonitor.js
```diff
@@ -10,276 +10,144 @@ const WorkspaceMonitor = new Lang.Class({
...
-    _fullscreenChanged: function() {
-        let primaryMonitor = Main.layoutManager.primaryMonitor;
-        let inFullscreen = primaryMonitor && primaryMonitor.inFullscreen;
-
-        if (this._inFullscreen != inFullscreen) {
-            this._inFullscreen = inFullscreen;
-            this._updateOverview();
-        }
-    },
+    _appHasVisibleWindows: function(app) {
```
cosimoc: I wonder if this shouldn't be a method of `ShellApp` itself, since it uses nothing from the workspace monitor.

js/ui/workspaceMonitor.js
```diff
@@ -10,276 +10,144 @@ const WorkspaceMonitor = new Lang.Class({
...
+    _onAppWindowsChanged: function(app) {
+        this._setAppVisible(app, this._appHasVisibleWindows(app));
 
-        // If the window was minimized remove it
-        if (this._windowIsMinimized(metaWindow)) {
-            this._removeMinimizedWindow(metaWindow);
+        // We need to track the stopped state here (instead of in _onAppStateChange),
+        // and independently from setting an app visible, because there are apps
+        // that go into "running" -> "stopped" -> "running" states because of the
+        // way they show their windows. This is the case of the Maps (marble) app.
```
cosimoc: The following code block implies that it's possible for an app to end up in a `STOPPED` state with one window still being visible. I think that is a bug in the `ShellWindowTracker`/`ShellAppSystem`/`ShellApp` machinery. It should never be possible that the app is stopped but a window is still visible, and we should not work that around here.

js/ui/workspaceMonitor.js
```diff
@@ -10,276 +10,144 @@ const WorkspaceMonitor = new Lang.Class({
...
-        if (!this._windowIsKnown(actor.meta_window)) {
-            return;
-        }
-
-        this._windowGoingInvisible();
-
-        // We'll show the overview when the destroy animation ends
-        this._addMinimizedWindow(actor.meta_window);
+    _setAppVisible: function(app, visible) {
+        if (visible)
```
cosimoc: Coding style: use curly braces even for one line blocks (multiple occurrences)

js/ui/workspaceMonitor.js
```diff
@@ -10,276 +10,144 @@ const WorkspaceMonitor = new Lang.Class({
...
-        if (!this._windowIsKnown(actor.meta_window)) {
-            return;
-        }
-
-        this._windowGoingInvisible();
-
-        // We'll show the overview when the destroy animation ends
-        this._addDestroyedWindow(actor.meta_window);
+    _trackApp: function(app) {
+        if (!(app in this._trackedApps))
```
cosimoc: Nitpick: prefer early return and inverting condition

js/ui/workspaceMonitor.js
```diff
@@ -10,276 +10,144 @@ const WorkspaceMonitor = new Lang.Class({
...
-            this._updateOverview();
-        }
-    },
+    _appHasVisibleWindows: function(app) {
+        let windows = app.get_windows();
+        for (let window of windows) {
+            // We do not count transient windows because of an issue with Audacity
+            // where a transient window was always being counted as visible even
+            // though it was minimized
+            if (window.get_transient_for())
```
cosimoc: Interesting, why were we not hitting the same problem before? I wonder if this couldn't cause other issues though when done generically.

joaquimrocha:
Closed as there are a new PR for this issue.

PR #117, opened 2016-02-17
issue/5950

GeorgesStavracas:
The Shell's Overview sets up a clone of the app grid to apply
the Saturation effect when minimized. The problem with this
approach, however, is that we end up never updating the real
child when something changes while we're not in Overview.

This wouldn't be a problem, but there may be unmaximized windows
that exposes part of the Overview - in this case, this issue
is pretty noticeable.

Fix that by toggling the Overview on and off.

[endlessm/eos-shell#5950]

### DIFF
--- a/js/ui/layout.js
+++ b/js/ui/layout.js
@@ -27,6 +27,9 @@ const DEFAULT_BACKGROUND_COLOR = Clutter.Color.from_pixel(0x2e3436ff);
 // Gsettings key to enable the message tray pressure barrier.
 const ENABLE_MESSAGE_TRAY_BARRIER_KEY = 'enable-message-tray-barrier'
 
+// GSettings key to track the text scaling factor
+const KEY_TEXT_SCALING_FACTOR = 'text-scaling-factor';
+
 // The message tray takes this much pressure
 // in the pressure barrier at once to release it.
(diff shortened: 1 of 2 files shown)


js/ui/layout.js
```diff
@@ -421,6 +424,22 @@ const LayoutManager = new Lang.Class({
...
+         * some random font rendering issues in apps, toggle the Overview.
+         */
+        Main.interfaceSettings.connect('changed::' + KEY_TEXT_SCALING_FACTOR, Lang.bind(this, function() {
+            if (!this._inOverview) {
+                Main.layoutManager.showOverview();
+
+                GLib.idle_add(GLib.PRIORITY_DEFAULT, Lang.bind(this, function() {
+                                  this.hideOverview();
+                                  return false;
+                              }));
```
smspillaz: Style: I believe that the end curly bracket here should align with `GLib`

smspillaz: Hmmm, showing the overview and then hiding it again immediately on a setting change seems suspect to me. Can't we just apply this setting change once the overview is next toggled?

cosimoc:
We used a different fix for this. The PR can be closed (and should not be merged).

PR #118, opened 2016-02-19
loginDialog: Fix gjs warning

mcatanzaro:
JS WARNING: [resource:///org/gnome/shell/gdm/loginDialog.js 1174]:
anonymous function does not always return a value

No behavior change, just silencing the warning.

[endlessm/eos-shell#6352]

### DIFF
--- a/js/gdm/loginDialog.js
+++ b/js/gdm/loginDialog.js
@@ -1179,6 +1179,7 @@ const LoginDialog = new Lang.Class({
                              this._promptEntry.reactive = false;
                              return this._beginVerificationForUser(userName);
                          }
+                         return undefined;
                      }];
 
         let batch = new Batch.ConsecutiveBatch(this, tasks);


js/gdm/loginDialog.js
```diff
@@ -1179,6 +1179,7 @@ const LoginDialog = new Lang.Class({
                              this._promptEntry.reactive = false;
                              return this._beginVerificationForUser(userName);
                          }
+                         return undefined;
```
mariospr: Do we really need to return something in this function? Couldn't we simply drop the other return when calling `_beginVerificationForUser()`? I think the return value is not being used anyway...

mcatanzaro: Yes, we really need it, that ensures the function doesn't return until the hold returned by `_beginVerificationForUser` has been canceled. The `ConsecutiveBatch` object ensures this. It's frankly really confusing and most uses of it have since been removed upstream, but it's very hard to get rid of without rebasing on the latest upstream code.

mariospr: Thanks for the clarification, I agree with you now then. Merging...

PR #121, opened 2016-02-24
Shell/6318

mcatanzaro:
[endlessm/eos-shell#6318]


mcatanzaro:
Warning: I just discovered there is a bug in the current version of eos-shell that causes the shell to not start if the hot corner is disabled. If you test this PR, leave the hot corner disabled, then revert back to the previous version of eos-shell, it will not start anymore, so be sure to enable the hot corner before reverting back. That bug happens to be fixed by the second commit in this PR.

rshuler:
Thanks for the warning :-)

PR #123, opened 2016-02-26
ShellApp: don't transition to STARTING state from launch

cosimoc:
Instead, track the number of speedwagon windows, in addition to the
regular logic, to decide if we should move to the STARTING state.
At the same time, make it possible to transition from STARTING to
STOPPED following the same logic.

[endlessm/eos-shell#6018]

### DIFF
--- a/src/shell-app.c
+++ b/src/shell-app.c
@@ -45,6 +45,7 @@ typedef struct {
   GSList *windows;
 
   guint interesting_windows;
+  guint speedwagon_windows;
 
   /* Whether or not we need to resort the windows; this is done on demand */
   gboolean window_sort_stale : 1;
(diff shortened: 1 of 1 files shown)


magcius:
*sigh*. Another round of hacks. We can never get this code right, can we.

I wonder if we should sit down and figure out what we want out of an API, rather than poking it with a stick in hope it doesn't fall over.

cosimoc:
@magcius I did sit down with @joaquimrocha and we thought about how this should work pretty thoroughly for endlessm/eos-shell#4190. This commit actually fixes a regression that was introduced in that story, which also happened to partially help with endlessm/eos-shell#6018

PR #124, opened 2016-03-01
appActivation: Prevent showing windows before speedwagon

joaquimrocha:
When the overview is hidden, the actors corresponding to visible
(non-minimized) windows are shown. As a result, the speedwagon's
animation does show up as smooth as it should.

These changes hide the mentioned actors while the speedwagon is shown
and show them again just before the speedwagon disappears. This way the
speedwagon's animation is smooth and there is no impact on the behavior
after it is gone.

https://phabricator.endlessm.com/T5419

### DIFF
--- a/js/ui/appActivation.js
+++ b/js/ui/appActivation.js
@@ -41,6 +41,8 @@ const AppActivationContext = new Lang.Class({
         this._timeoutId = 0;
 
         this._appActivationTime = 0;
+
+        this._hiddenWindows = [];
     },
 
     _doActivate: function(showSplash, timestamp) {
@@ -97,6 +99,12 @@ const AppActivationContext = new Lang.Class({
             return;
         }
 
+        // Prevent windows from being shown when the overview is hidden so it does
+        // not affect the speedwagon's animation
+        if (Main.overview.visible) {
+            this._hideWindows();
+        }
+
         this._cancelled = false;
         this._splash = new SpeedwagonSplash(this._app);
         this._splash.connect('close-clicked', Lang.bind(this, function() {
@@ -130,6 +138,8 @@ const AppActivationContext = new Lang.Class({
     },
 
     _clearSplash: function() {
+        this._resetWindowsVisibility();
+
         if (this._splash) {
             this._splash.rampOut();
             this._splash = null;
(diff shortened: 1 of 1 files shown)


js/ui/appActivation.js
```diff
@@ -139,8 +145,10 @@ const AppActivationContext = new Lang.Class({
     _maybeClearSplash: function() {
         // Clear the splash only when we've waited at least 700ms,
         // and when the app has transitioned to the running state...
-        if (this._appStateId == 0 && this._timeoutId == 0)
+        if (this._appStateId == 0 && this._timeoutId == 0) {
+            this._resetWindowsVisibility();
```
cosimoc: Should probably be done inside `_clearSplash()` as that's called from elsewhere too?

js/ui/appActivation.js
```diff
@@ -56,6 +58,10 @@ const AppActivationContext = new Lang.Class({
         }
 
         if (showSplash) {
+            if (Main.overview.visible) {
+                this._hideWindows();
```
cosimoc: Should probably be done inside `this.showSplash()`, and only if the logic there actually decides to show the splash.
But not when `showSplash()` is called from `DesktopClient`, as that is only called when one application launches another (e.g. a file manager opening a file)

js/ui/appActivation.js
```diff
@@ -150,6 +162,27 @@ const AppActivationContext = new Lang.Class({
...
+    _resetWindowsVisibility: function() {
+        for (let actor of this._hiddenWindows) {
+            actor.visible = true;
+        }
+
+        this._hiddenWindows = [];
+    },
+
+    _hideWindows: function() {
+        let windows = global.window_group.get_children();
```
cosimoc: You should use `shell_global_get_window_actors()`

js/ui/appActivation.js
cosimoc: I don't think you need this anymore if you do it from `_clearSplash()`

joaquimrocha: I'd swear I moved it... Maybe it just copied it but I'm still surprised it got through. (I always use add -p for reviewing the changes I add)

PR #125, opened 2016-03-03
Endlessm/t6387

ebassi:
https://phabricator.endlessm.com/T6387


src/shell-app-system.c
```diff
@@ -87,6 +94,31 @@ static void shell_app_system_class_init(ShellAppSystemClass *klass)
 }
 
 static void
+scan_alias_to_id (ShellAppSystem *self)
```
cosimoc: This function looks unused

ebassi: Yes, I'm an idiot and didn't push the fixed up commit that called this.

Fixed.

PR #129, opened 2016-03-11
T10595

ptomato:
https://phabricator.endlessm.com/T10595

### DIFF
--- a/js/ui/search.js
+++ b/js/ui/search.js
@@ -174,7 +174,7 @@ const ListSearchResult = new Lang.Class({
                                y_align: St.Align.MIDDLE });
 
         let title = new St.Label({ style_class: 'list-search-result-title',
-                                   text: this.metaInfo['name'],
+                                   text: this.metaInfo['name'] || '',
                                    y_align: Clutter.ActorAlign.CENTER });
         details.add(title, { x_fill: false,
                              y_fill: true,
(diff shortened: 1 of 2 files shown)


js/ui/search.js
```diff
@@ -174,7 +174,7 @@ const ListSearchResult = new Lang.Class({
                                y_align: St.Align.MIDDLE });
 
         let title = new St.Label({ style_class: 'list-search-result-title',
-                                   text: this.metaInfo['name'],
+                                   text: this.metaInfo['name'] || '',
```
magcius: WTF? No.

cosimoc: Yeah, I debated merging this too... The other option is to just filter out meta infos with no name. The shell definitely should not throw exceptions or crash when the provider sends malformed results.
@magcius Would you prefer the other approach?

magcius: I don't think anything major should happen, since I'm quite sure all result construction is within a try/catch. So you'll see errors, but they'll be properly caught. I think more than anything all the results for that app would be dead, which is a bit heavyhanded of a punishment. The logic is "if the app gave us any junk at all, it can't be trusted properly".

I don't like this option, because the user will just see an empty bubble as a result.

cosimoc: @magcius I am not against the punishment for apps that give us garbage, but it does not look like the result construction is within a try/catch to me.

PR #131, opened 2016-03-17
Never crash when launching eos-tutorial

mcatanzaro:
https://phabricator.endlessm.com/T3727

PR #132, opened 2016-03-21
workspace: Do not crash when launching eos-tutorial

mcatanzaro:
We cannot compute window positions before the Workspace actor has been
mapped.

https://phabricator.endlessm.com/T3881

(Same as https://github.com/endlessm/eos-desktop/pull/131)


mcatanzaro:
As Roddy pointed out, this does not fix T3881.

PR #135, opened 2016-03-25
WindowManager: Prevent error when sliding out a side component

joaquimrocha:
The error happens when the MetaWindow of the give actor in
_slideSideComponentOut may be null.

This may not solve the whole issue, please see my comments in the task.

https://phabricator.endlessm.com/T10999


mcatanzaro:
This can be closed now, right @joaquimrocha?

joaquimrocha:
Yes. Thanks.

PR #139, opened 2016-04-07
T6221

cosimoc:
https://phabricator.endlessm.com/T6221

### DIFF
--- a/js/ui/screenShield.js
+++ b/js/ui/screenShield.js
@@ -226,7 +226,7 @@ const ScreenShield = new Lang.Class({
     },
 
     _onLightboxShown: function() {
-        this.activate(false);
+        this.emit('active-changed');
     },
 
     showDialog: function() {
@@ -322,7 +322,7 @@ const ScreenShield = new Lang.Class({
 
     get active() {
         // See the comment on the activate method.
-        return this._isLocked;
+        return this._lightbox.shown || this._isLocked;
     },
 
     get activationTime() {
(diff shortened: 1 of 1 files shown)


js/ui/screenShield.js
```diff
@@ -402,17 +402,7 @@ const ScreenShield = new Lang.Class({
...
-        // but now we do that from lockScreenShown, which means
-        // there is a 0.3 seconds window during which the lock
-        // screen is effectively visible and the screen is locked, but
-        // the DBus interface reports the screensaver is off.
-        // This is because when we emit ActiveChanged(true),
-        // gnome-settings-daemon blanks the screen, and we don't want
-        // blank during the animation.
-        // This is not a problem for the idle fade case, because we
-        // activate without animation in that case.
+        this.emit('active-changed');
```
mcatanzaro: This looks good. I thought we needed to emit active-changed here, but got confused by the stale comment.

mcatanzaro:
I will look at the last commit a bit closer tomorrow before merging this.

mcatanzaro:
I trust this is fine, but since this is not urgent, let me play with the code a bit more over the weekend to see if it's easy to make this cleaner/safer. I'm not really comfortable with having only partially merged together the activated/locked states.

mcatanzaro:
OK, let's go with this, since you've tested it and it's minimally-intrusive... 'active-changed' doesn't blank the screen in my VM for some reason I haven't determined, so I'm having trouble working on this, unfortunately.

PR #140, opened 2016-04-14
viewSelector: redraw the clones on large text setting changes

GeorgesStavracas:
This commit is a workaround on the Clutter misbehavior with
Framebuffer Objects. Precisely, the clone won't update properly
because the original actor mistakenly doesn't detect the living
clone and doesn't update correctly.

Fix that by adding a redraw'ing routine in an idle, to give
time for the original actor to update.

https://phabricator.endlessm.com/T5950

### DIFF
--- a/js/ui/viewSelector.js
+++ b/js/ui/viewSelector.js
@@ -391,6 +391,8 @@ const ViewsClone = new Lang.Class({
     Extends: St.Widget,
 
     _init: function(viewSelector, viewsDisplay, forOverview) {
+        let settings = Clutter.Settings.get_default();
+
         this._viewSelector = viewSelector;
         this._viewsDisplay = viewsDisplay;
         this._forOverview = forOverview;
@@ -444,6 +446,25 @@ const ViewsClone = new Lang.Class({
                                    time: OverviewControls.SIDE_CONTROLS_ANIMATION_TIME,
                                    transition: 'easeOutQuad' });
             }
+        }));
+
+        settings.connect('notify::font-dpi', Lang.bind(this, function() {
+            let overviewVisible = Main.layoutManager.overviewGroup.visible;
+            let saturationEnabled = this._saturation.enabled;
+
+            /* Maybe because of the already known issue with FBO and ClutterClones,
+             * simply redrawing the overview group without assuring it is visible
+             * won't work. Clutter was supposed to do that, but it doesn't. The
+             * FBO, in this case, is introduced through the saturation effect.
+             */
+            this._saturation.enabled = false;
+            Main.layoutManager.overviewGroup.visible = true;
+
+            Main.layoutManager.overviewGroup.queue_redraw();
+
+            /* Restore the previous states */
+            Main.layoutManager.overviewGroup.visible = overviewVisible;
+            this._saturation.enabled = saturationEnabled;
         }));
     },
 


js/ui/viewSelector.js
```diff
@@ -394,6 +397,7 @@ const ViewsClone = new Lang.Class({
         this._viewSelector = viewSelector;
         this._viewsDisplay = viewsDisplay;
         this._forOverview = forOverview;
+        this._settings = new Gio.Settings({ schema: DESKTOP_INTERFACE_SCHEMA });
```
cosimoc: Nitpick: there should be no need to keep a reference to `this._settings`.

js/ui/viewSelector.js
```diff
@@ -445,6 +449,33 @@ const ViewsClone = new Lang.Class({
...
+                 * simply redrawing the overview group without assuring it is
+                 * visible won't work. Clutter was supposed to do that, but it
+                 * doesn't.
+                 */
+                Main.layoutManager.overviewGroup.visible = true;
+                Main.layoutManager.overviewGroup.queue_redraw();
+
+                /* Restore the previous states */
+                Main.layoutManager.overviewGroup.visible = overviewVisible;
+                this._saturation.enabled = saturationEnabled;
```
cosimoc: Should return `false` here.

js/ui/viewSelector.js
```diff
@@ -445,6 +449,33 @@ const ViewsClone = new Lang.Class({
...
+             * Add it in an idle to give Clutter some time to
+             * recalculate whatetever it needs to. Simply queueing
+             * redraw right after the setting changes doesn't
+             * work.
+             */
+            GLib.idle_add(GLib.PRIORITY_DEFAULT, Lang.bind(this, function() {
+                let overviewVisible = Main.layoutManager.overviewGroup.visible;
+                let saturationEnabled = this._saturation.enabled;
+
+                this._saturation.enabled = false;
```
cosimoc: Are you disabling the saturation for the same reason in the comment below? If so, please move the line after the comment, otherwise it probably would deserve a separate comment.

js/ui/viewSelector.js
GeorgesStavracas: Done

js/ui/viewSelector.js
GeorgesStavracas: This does not seem to work. I think GJS doesn't keep a reference to it anyhow.

js/ui/viewSelector.js
```diff
@@ -445,6 +449,37 @@ const ViewsClone = new Lang.Class({
                                    transition: 'easeOutQuad' });
             }
         }));
+
+        this._settings.connect('changed::' + KEY_TEXT_SCALING_FACTOR, Lang.bind(this, function() {
+            /*
+             * Add it in an idle to give Clutter some time to
+             * recalculate whatetever it needs to. Simply queueing
+             * redraw right after the setting changes doesn't
+             * work.
```
cosimoc: Sorry for the back and forth on this, but I believe I understand the reason you can't just queue a redraw right away here.
Clutter does not listen to this gsetting directly; however it listens to the `Xft/DPI` and `Gdk/UnscaledDPI` xsettings, which gnome-settings-daemon is responsible of updating whenever the gsetting changes.

In other words, you should be able to connect to notifications on the `font-dpi` property of `ClutterSettings` and avoid the idle callback.

js/ui/viewSelector.js
cosimoc: Very odd! But may not be necessary after all - see my comment below.

js/ui/viewSelector.js
```diff
@@ -445,6 +447,27 @@ const ViewsClone = new Lang.Class({
...
+            this._saturation.enabled = false;
+            Main.layoutManager.overviewGroup.visible = true;
+
+            Main.layoutManager.overviewGroup.queue_redraw();
+
+            /* Restore the previous states */
+            Main.layoutManager.overviewGroup.visible = overviewVisible;
+            this._saturation.enabled = saturationEnabled;
+
+            return GLib.SOURCE_REMOVE;
```
ebassi: There's no need to return anything, here; the `GObject::notify` signal does not have a return value.

PR #146, opened 2016-05-13
T11618

cosimoc:
https://phabricator.endlessm.com/T11618

### DIFF
--- a/js/ui/internetSearch.js
+++ b/js/ui/internetSearch.js
@@ -2,6 +2,7 @@
 
 const GLib = imports.gi.GLib;
 const Gio = imports.gi.Gio;
+const Json = imports.gi.Json;
 const Lang = imports.lang;
 
 const Main = imports.ui.main;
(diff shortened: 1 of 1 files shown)


js/ui/internetSearch.js
```diff
@@ -69,7 +69,7 @@ const InternetSearchProvider = new Lang.Class({
                 name = _("Open \"%s\" in browser").format(uri);
             } else if (resultId.startsWith('search:')) {
                 let query = resultId.slice('search:'.length);
-                name = _("Search Google for \"%s\"").format(query);
+                name = _("Search the web for \"%s\"").format(query);
```
rshuler: I'm not sure if we want this string change.  In the past I recall Matt being pretty specific about how he wanted to present the search results.  Another easy option for now would be to leave "Google" in the English string and simply edit the Chinese translation to replace "Google" with "Baidu".

rshuler: Oh, I see the later commit that changes this.

js/ui/internetSearch.js
```diff
@@ -54,22 +90,31 @@ const InternetSearchProvider = new Lang.Class({
...
-                name = _("Search the web for \"%s\"").format(query);
+                let engineName = this._getEngineName();
+
+                if (engineName) {
+                    /* Translators: the first %s is the search engine name, and the second
+                     * is the search string. For instance, 'Search Google for "hello"'.
+                     */
+                    name = _("Search %s for \"%s\"").format(engineName, query);
+                } else {
+                    name = _("Search the web for \"%s\"").format(query);
```
rshuler: Another possible way to do this is use the first format string with engineName = _("the web"), which reduces the total number of words to translate.  Also, I think we should use "the internet" rather than "the web" for consistency with how we describe this in the help center:
> The global search bar makes search easy! You can look for apps, documents, photos and more, right from your desktop. If you have an internet conection, you can even look for things that are on the internet, like you would on google, with the global search tool.

js/ui/internetSearch.js
```diff
@@ -42,9 +43,44 @@ const InternetSearchProvider = new Lang.Class({
...
+        try {
+            parser.load_from_file(path);
+        } catch (e) {
+            logError(e, 'error while parsing Chromium preferences');
+            return null;
+        }
+
+        let root = parser.get_root().get_object();
+        let searchProviderData = root.get_object_member('default_search_provider_data');
+        if (!searchProviderData)
```
rshuler: Javascript, so use curly braces even if just single statement

js/ui/internetSearch.js
```diff
@@ -42,9 +43,44 @@ const InternetSearchProvider = new Lang.Class({
...
+            return null;
+
+        let templateUrlData = searchProviderData.get_object_member('template_url_data');
+        if (!templateUrlData)
```
rshuler: Ditto

js/ui/internetSearch.js
cosimoc: I will change this to say "the internet", but we can't really build the string as you propose. Some languages will need to translate "Search Google for %s" and "Search the internet for %s" as two separate strings, and that's why I did it this way.

rshuler: I'm surprised that any language would change the translation of "Search %s for \"%\"s" depending on whether the first "%s" is "Google" or "the internet".  Not a big deal either way, but I'm curious to understand why so I can keep that in mind in the future.  We can talk offline.

ptomato: In Dutch, for example, the preposition would change — you search "through" Google, but "on" the internet for something.

rshuler: Ah, thank you @ptomato.  That makes sense now.

PR #147, opened 2016-05-13
panel: Show/hide the social bar icon depending on a gsetting

joaquimrocha:
https://phabricator.endlessm.com/T11619


js/ui/panel.js
```diff
@@ -412,6 +416,27 @@ const Panel = new Lang.Class({
                 });
             })
         );
+
+        global.settings.connect('changed::' + SETTINGS_ENABLE_SOCIAL_BAR_KEY,
+                                Lang.bind(this, this._updateSocialBar));
```
cosimoc: Instead of having this in `panel.js`, you should move that logic in `status/social.js` itself. You can take inspiration from the accessibility indicator: https://github.com/endlessm/eos-desktop/blob/master/js/ui/status/accessibility.js#L83

PR #149, opened 2016-05-27
T11713

cosimoc:
https://phabricator.endlessm.com/T11713

### DIFF
--- a/js/ui/status/keyboard.js
+++ b/js/ui/status/keyboard.js
@@ -261,9 +261,9 @@ const InputSourceSessionSettings = new Lang.Class({
         this._settings.connect('changed::' + this._KEY_PER_WINDOW, Lang.bind(this, this._emitPerWindowChanged));
     },
 
-    get inputSources() {
+    _getSourcesList: function(key) {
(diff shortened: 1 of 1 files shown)


js/ui/status/keyboard.js
```diff
@@ -453,6 +493,27 @@ const InputSourceManager = new Lang.Class({
...
+            let mruSettings = this._settings.mruSources;
+            for (let i = 0; i < mruSettings.length; i++) {
+                let mruSettingSource = mruSettings[i];
+                let mruSource = null;
+
+                for (let j = 0; j < sourcesList.length; j++) {
+                    let source = sourcesList[j];
+                    if (source.type == mruSettingSource.type &&
+                        source.id == mruSettingSource.id) {
+                        mruSource = source;
```
joaquimrocha: Why not pushing *source* to the list directly here? mruSource doesn't seem to be used anywhere else.

cosimoc: That's true... But since this was pushed upstream like this already, I prefer merging it as-is to avoid carrying an additional delta.

PR #150, opened 2016-05-30
T11422

joaquimrocha:
https://phabricator.endlessm.com/T11422

### DIFF
--- a/js/ui/iconGridLayout.js
+++ b/js/ui/iconGridLayout.js
@@ -234,8 +234,10 @@ const IconGridLayout = new Lang.Class({
 
         let filename = info.get_filename();
         let userDir = GLib.get_user_data_dir();
-        if (filename && userDir && GLib.str_has_prefix(filename, userDir)) {
-            // only delete .desktop files in the user's local data folder
+        if (filename && userDir && GLib.str_has_prefix(filename, userDir) &&
+            info.get_string('X-Endless-CreatedBy') === 'eos-desktop') {
+            // only delete .desktop files if they are in the user's local data
+            // folder and they were created by eos-desktop
             info.delete();
         }
     },
--- a/src/shell-app.c
+++ b/src/shell-app.c
@@ -1401,14 +1401,6 @@ shell_app_create_custom_launcher_with_name (ShellApp *app,
                          G_KEY_FILE_DESKTOP_KEY_NAME,
                          label);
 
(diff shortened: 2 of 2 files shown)


src/shell-app.c
```diff
@@ -1428,6 +1420,24 @@ shell_app_create_custom_launcher_with_name (ShellApp *app,
 
   new_path = g_build_filename (g_get_user_data_dir (), "applications", shell_app_get_id (app), NULL);
 
+  if (!g_file_test (new_path, G_FILE_TEST_EXISTS))
+    {
+      /* Create a new 'X-Endless-CreatedBy' key to indicate the file was created
+       * by eos-desktop */
+      g_key_file_set_string (keyfile,
```
cosimoc: I think it should be fine to set this even when the file already exists. Otherwise if someone starts with a custom desktop file from a version that didn't have the marker, and then renamed the file again, they won't be able to remove it.

js/ui/iconGridLayout.js
```diff
@@ -234,8 +234,10 @@ const IconGridLayout = new Lang.Class({
 
         let filename = info.get_filename();
         let userDir = GLib.get_user_data_dir();
-        if (filename && userDir && GLib.str_has_prefix(filename, userDir)) {
-            // only delete .desktop files in the user's local data folder
+        if (filename && userDir && GLib.str_has_prefix(filename, userDir) &&
+            info.get_string('X-Endless-CreatedBy') === 'eos-desktop') {
```
cosimoc: You could make the `X-Endless-CreatedBy` a public define in `shell-app.h` and use it from here and `shell-app.c` in the previous commit.

src/shell-app.c
joaquimrocha: That defeats the purpose of the patch because I noticed this problem initially when using *xdg-app* formerly on the user folder. It installed desktop files into *~/.local/share/applications* and upon removing the files from the desktop grid, they would be deleted from the folder too, losing the desktop file.
So that's why we need this fix mainly. An alternative may be to use a different (not so standard) folder to hold the renamed desktop files, in which case we can even drop the special key.

cosimoc: I see your point. Looks fine then, since even if the desktop file would not be deleted in the case I mention, it would still be removed from the icon grid (which is what counts).

PR #155, opened 2016-08-05
Mission: Add wobbly windows.

smspillaz:
Link to libwindowfx_wobbly (from libwobbly) and add a new
EndlessShellFXWobbly class to bridge between a clutter effect
and the wobbly::Model/wobbly::Anchor.

Use endless_shell_fx_wobbly_grab to grab a point on the mesh
and move it around.

Use endless_shell_fx_wobbly_ungrab to release a point on the mesh.

Each of the above functions may only be called once whilst in the
opposing state and cannot be called again whilst in the same state.

EOSShellWobbly is a JavaScript subclass of EndlessShellFXWobbly
which handles integration between the shell itself and the wobbly
effect. It provides API to handle mouse grab / ungrab of a window
and window movement whilst grabbed.

Provide the following new keys in org.gnome.shell:

 'wobbly-spring-k' : Spring K
 'wobbly-spring-friction' : Spring Friction
 'wobbly-slowdown-factor' : Amount to slow effect down
 'wobbly-object-movement-range' : Amount points on the mesh are
  allowed to deviate from their target position once the mesh is
  settled.

https://phabricator.endlessm.com/T1790

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -114,7 +114,7 @@ PKG_CHECK_MODULES(EOS_SHELL, gio-unix-2.0 >= $GIO_MIN_VERSION
 
 PKG_CHECK_MODULES(EOS_SHELL_JS, gio-2.0 gjs-internals-1.0 >= $GJS_MIN_VERSION)
 PKG_CHECK_MODULES(ST, clutter-1.0 gtk+-3.0 libcroco-0.6 >= 0.6.8 x11)
-PKG_CHECK_MODULES(EOS_SHELL_FX, clutter-1.0 >= 1.16 glib-2.0 >= 2.38 gdk-pixbuf-2.0 libwindowfx_wobbly)
+PKG_CHECK_MODULES(EOS_SHELL_FX, clutter-1.0 >= 1.16 glib-2.0 >= 2.38 libwindowfx_wobbly)
 PKG_CHECK_MODULES(SHELL_PERF_HELPER, gtk+-3.0 gio-2.0)
 PKG_CHECK_MODULES(SHELL_HOTPLUG_SNIFFER, gio-2.0 gdk-pixbuf-2.0)
 PKG_CHECK_MODULES(TRAY, gtk+-3.0)
--- a/configure.ac
+++ b/configure.ac
@@ -114,6 +114,7 @@ PKG_CHECK_MODULES(EOS_SHELL, gio-unix-2.0 >= $GIO_MIN_VERSION
 
 PKG_CHECK_MODULES(EOS_SHELL_JS, gio-2.0 gjs-internals-1.0 >= $GJS_MIN_VERSION)
 PKG_CHECK_MODULES(ST, clutter-1.0 gtk+-3.0 libcroco-0.6 >= 0.6.8 x11)
+PKG_CHECK_MODULES(EOS_SHELL_FX, clutter-1.0 >= 1.16 glib-2.0 >= 2.38 gdk-pixbuf-2.0 libwindowfx_wobbly)
 PKG_CHECK_MODULES(SHELL_PERF_HELPER, gtk+-3.0 gio-2.0)
 PKG_CHECK_MODULES(SHELL_HOTPLUG_SNIFFER, gio-2.0 gdk-pixbuf-2.0)
 PKG_CHECK_MODULES(TRAY, gtk+-3.0)
--- a/js/ui/windowManager.js
+++ b/js/ui/windowManager.js
@@ -140,14 +140,6 @@ const DisplayChangeDialog = new Lang.Class({
 const EOSShellWobbly = new Lang.Class({
     Name: 'EOSShellWobbly',
     Extends: EndlessShellFX.Wobbly,
-    Properties: {
-        'settings' : GObject.ParamSpec.object('settings',
-                                              'Effect GSettings',
-                                              'Wobbly Effect GSettings',
-                                              GObject.ParamFlags.CONSTRUCT_ONLY |
-                                              GObject.ParamFlags.WRITABLE,
-                                              Gio.Settings)
-    },
 
     _init: function(params) {
         this.parent(params);
--- a/js/ui/windowManager.js
+++ b/js/ui/windowManager.js
@@ -1629,7 +1629,6 @@ const WindowManager = new Lang.Class({
         if (!window)
             return;
 
-
         let actor = window.get_compositor_private();
         let effect = actor.get_effect('endless-wobbly');
 
--- a/js/ui/windowManager.js
+++ b/js/ui/windowManager.js
@@ -173,7 +173,6 @@ const EOSShellWobbly = new Lang.Class({
     },
 
     grabbedByMouse: function() {
-
         if (!global.get_settings().get_boolean('wobbly-effect'))
             return;
 
--- a/js/ui/windowManager.js
+++ b/js/ui/windowManager.js
@@ -182,7 +182,7 @@ const EOSShellWobbly = new Lang.Class({
 
         this._lastPosition = actor.get_position();
         this._positionChangedId =
-            actor.connect('allocation-changed', Lang.bind(this, function (actor) {
+            actor.connect('notify::position', Lang.bind(this, function (actor) {
                 let position = actor.get_position();
                 let dx = position[0] - this._lastPosition[0];
                 let dy = position[1] - this._lastPosition[1];
--- a/js/ui/windowManager.js
+++ b/js/ui/windowManager.js
@@ -155,7 +155,7 @@ const EOSShellWobbly = new Lang.Class({
     },
 
     set settings (val) {
-         if (val['schema-id'] !== 'org.gnome.shell')
+         if (val.schema_id !== 'org.gnome.shell')
              throw Error('EOSShellWobbly: Injected schema must be of ' +
                          'org.gnome.shell');
 
--- a/js/ui/windowManager.js
+++ b/js/ui/windowManager.js
@@ -1604,7 +1604,6 @@ const WindowManager = new Lang.Class({
     },
 
     _windowGrabbed: function(display, screen, window, op) {
-
         // Occassionally, window can be null, in cases where grab-op-begin
         // was emitted on a window from shell-toolkit. Ignore these grabs.
         if (!window)
--- a/js/ui/windowManager.js
+++ b/js/ui/windowManager.js
@@ -161,8 +161,8 @@ const EOSShellWobbly = new Lang.Class({
 
          this._settings = val;
 
-         let binder = Lang.bind (this, function(key, prop) {
-             this._settings.bind (key, this, prop, Gio.SettingsBindFlags.GET);
+         let binder = Lang.bind(this, function(key, prop) {
+             this._settings.bind(key, this, prop, Gio.SettingsBindFlags.GET);
          });
 
          /* Bind to effect properties */
--- a/js/ui/windowManager.js
+++ b/js/ui/windowManager.js
@@ -1605,7 +1605,6 @@ const WindowManager = new Lang.Class({
     },
 
     _windowCanWobble: function(window, op) {
-
         if (window.is_override_redirect() ||
             op != Meta.GrabOp.MOVING)
             return false;
--- a/js/ui/windowManager.js
+++ b/js/ui/windowManager.js
@@ -5,6 +5,8 @@ const Clutter = imports.gi.Clutter;
 const Gdk = imports.gi.Gdk;
 const GLib = imports.gi.GLib;
 const Gio = imports.gi.Gio;
+const GObject = imports.gi.GObject;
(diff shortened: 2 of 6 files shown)


configure.ac
```diff
@@ -114,6 +114,7 @@ PKG_CHECK_MODULES(EOS_SHELL, gio-unix-2.0 >= $GIO_MIN_VERSION
 
 PKG_CHECK_MODULES(EOS_SHELL_JS, gio-2.0 gjs-internals-1.0 >= $GJS_MIN_VERSION)
 PKG_CHECK_MODULES(ST, clutter-1.0 gtk+-3.0 libcroco-0.6 >= 0.6.8 x11)
+PKG_CHECK_MODULES(EOS_SHELL_FX, clutter-1.0 >= 1.16 glib-2.0 >= 2.38 gdk-pixbuf-2.0 libwindowfx_wobbly)
```
cosimoc: I don't think `gdk-pixbuf-2.0` is needed here.

js/ui/windowManager.js
```diff
@@ -135,6 +137,76 @@ const DisplayChangeDialog = new Lang.Class({
...
+                                              Gio.Settings)
+    },
+
+    _init: function(params) {
+        this.parent(params);
+        this._settings = null;
+    },
+
+    set settings (val)
+    {
```
cosimoc: Nitpick: curly brace on same line

js/ui/windowManager.js
```diff
@@ -135,6 +137,76 @@ const DisplayChangeDialog = new Lang.Class({
...
+         if (val['schema-id'] !== 'org.gnome.shell')
```
cosimoc: It's more idiomatic to use `val.schema_id`

js/ui/windowManager.js
```diff
@@ -135,6 +137,76 @@ const DisplayChangeDialog = new Lang.Class({
...
+             throw Error('EOSShellWobbly: Injected schema must be of ' +
+                         'org.gnome.shell');
+
+         this._settings = val;
+
+         let binder = Lang.bind (this, function(key, prop) {
```
cosimoc: Missing space before paren (multiple cases)

js/ui/windowManager.js
```diff
@@ -135,6 +137,76 @@ const DisplayChangeDialog = new Lang.Class({
...
+
+         /* Bind to effect properties */
+         binder('wobbly-spring-k', 'spring-k');
+         binder('wobbly-spring-friction', 'friction');
+         binder('wobbly-slowdown-factor', 'slowdown-factor');
+         binder('wobbly-object-movement-range', 'object-movement-range');
+    },
+
+    grabbedByMouse: function() {
+
```
cosimoc: Remove empty line

js/ui/windowManager.js
```diff
@@ -135,6 +137,76 @@ const DisplayChangeDialog = new Lang.Class({
...
+        if (!global.get_settings().get_boolean('wobbly-effect'))
+            return;
+
+        let position = global.get_pointer();
+        let actor = this.get_actor();
+        this.grab(position[0], position[1]);
+
+        this._lastPosition = actor.get_position();
+        this._positionChangedId =
+            actor.connect('allocation-changed', Lang.bind(this, function (actor) {
```
cosimoc: Perhaps you should connect to `notify::position` here instead of `allocation-changed`, since that's in reality what you're tracking?

js/ui/windowManager.js
```diff
@@ -1528,4 +1603,53 @@ const WindowManager = new Lang.Class({
...
+        if (!effect) {
+            effect = new EOSShellWobbly({ settings: global.get_settings() });
+            actor.add_effect_with_name('endless-wobbly', effect);
+        }
+
+        effect.grabbedByMouse();
+    },
+
+    _windowUngrabbed: function(display, op, window) {
+
```
cosimoc: Remove empty line

js/ui/windowManager.js
```diff
@@ -135,6 +137,76 @@ const DisplayChangeDialog = new Lang.Class({
...
+                this.move_by(dx, dy);
+                this._lastPosition = position;
+            }));
+    },
+
+    ungrabbedByMouse: function() {
+        // Only continue if we have an active grab and change notification
+        // on movement
+        if (this._positionChangedId !== undefined) {
```
cosimoc: Nitpick: prefer an early return

js/ui/windowManager.js
```diff
@@ -1528,4 +1603,53 @@ const WindowManager = new Lang.Class({
         let dialog = new DisplayChangeDialog(this._shellwm);
         dialog.open();
     },
+
+    _windowCanWobble: function(window, op) {
+
```
cosimoc: Remove empty line

src/Makefile.am
```diff
@@ -246,6 +246,22 @@ libeos_shell_js_la_CPPFLAGS =		\
 
 ########################################
 
+# Set up headers for libeos-shell-fx
+eos_shell_fxincludedir = $(includedir)
+eos_shell_fxinclude_HEADERS = wobbly-effect.h
+
+eos_shell_fx_cppflags = $(AM_CPPFLAGS) $(EOS_SHELL_FX_CFLAGS) -std=c++11 -Werror -Wall -Wextra -Wno-error=unused-parameter -Wno-error=deprecated-declarations
```
cosimoc: Do you really need to manually specify all these compiler flags here?

smspillaz: The philosophy here is mainly that they don't hurt and they catch bad things before they sneak into production code. I don't believe they're enabled for the rest of the project (nor would we want to do that, shell already compiles with warnings in a few places and now isn't the right time to go fixing all of those).

I suppose the tradeoff is that you have a slightly higher maintenance burden in terms of dealing with two different sets of warning flags. If there's a reason why that burden is high enough to justify having the same flags, then I can remove the specification, otherwise I think it should stay.

That said, I believe -std=c++11 is required for some of the functionality this code uses, or at least some of the stuff that's in libwobbly's headers. If we really cared, I could create a c++98 compatible interface, but since c++11 has been around for what feels like forever now, I'm not sure if there's really any benefit in doing so.

(Caveat: Back when I wrote this feature I was a much more of a C++ fanatic than I am now. Not that I'm jaded by the language, just that for whatever reason, I've had a lot more exposure to C, Python and JS lately).

js/ui/windowManager.js
smspillaz: An interesting point actually - I've noticed that it takes movement by 3-4 pixels for the effect to "kick in" and I'm wondering if this is the cause. I'm going to give `notify::position` a try now - perhaps it might fix that one last annoying glitch.

smspillaz: Oh wow, okay, its much smoother now :) Thanks so much for the suggestion! 👍

smspillaz:
All other comments addressed in review follow-up commits

js/ui/windowManager.js
```diff
@@ -135,6 +137,76 @@ const DisplayChangeDialog = new Lang.Class({
...
+         /* Bind to effect properties */
+         binder('wobbly-spring-k', 'spring-k');
+         binder('wobbly-spring-friction', 'friction');
+         binder('wobbly-slowdown-factor', 'slowdown-factor');
+         binder('wobbly-object-movement-range', 'object-movement-range');
+    },
+
+    grabbedByMouse: function() {
+        if (!global.get_settings().get_boolean('wobbly-effect'))
```
cosimoc: Should use `this._settings`.
Alternatively, you can remove the settings property entirely, and use `global.settings` everywhere -- actually I prefer that I think.

js/ui/windowManager.js
```diff
@@ -1528,4 +1603,51 @@ const WindowManager = new Lang.Class({
...
+    _windowCanWobble: function(window, op) {
+        if (window.is_override_redirect() ||
+            op != Meta.GrabOp.MOVING)
+            return false;
+
+        return true;
+    },
+
+    _windowGrabbed: function(display, screen, window, op) {
+
```
cosimoc: Remove empty line

js/ui/windowManager.js
```diff
@@ -1528,4 +1603,51 @@ const WindowManager = new Lang.Class({
...
+        effect.grabbedByMouse();
+    },
+
+    _windowUngrabbed: function(display, op, window) {
+        // Occassionally, window can be null, in cases where grab-op-end
+        // was emitted on a window from shell-toolkit. Ignore these grabs.
+        if (!window)
+            return;
+
```
cosimoc: Remove empty line

src/Makefile.am
cosimoc: I don't want one part of the project to use different flags than the rest... Please remove them, but keep the C++11 flag if needed. (Or otherwise, find a way to apply them to all the project instead, but I don't think it's worth investing in that direction given that upstream now does just that -- https://git.gnome.org/browse/gnome-shell/tree/configure.ac#n225)

src/wobbly-effect.cpp
```diff
@@ -0,0 +1,580 @@
...
+    EndlessShellFXWobbly *wobbly_effect =
+        ENDLESS_SHELL_FX_WOBBLY (effect);
+    EndlessShellFXWobblyPrivate *priv =
+        reinterpret_cast <EndlessShellFXWobblyPrivate *> (endless_shell_fx_wobbly_get_instance_private (wobbly_effect));
+
+
+    CLUTTER_EFFECT_CLASS (endless_shell_fx_wobbly_parent_class)->get_paint_volume (effect,
+                                                                                   volume);
+
+    g_return_val_if_fail (volume, FALSE);
```
cosimoc: I would not keep this here

src/wobbly-effect.cpp
```diff
@@ -0,0 +1,580 @@
...
+                             G_PARAM_WRITABLE);
+
+    g_object_class_install_properties (object_class, PROP_LAST, object_properties);
+}
+
+ClutterEffect *
+endless_shell_fx_wobbly_new ()
+{
+    return reinterpret_cast <ClutterEffect *> (g_object_new (ENDLESS_SHELL_FX_TYPE_WOBBLY,
+                                                             "x-tiles", 32,
```
cosimoc: These are the default values -- no need to repeat them.

src/wobbly-effect.cpp
```diff
@@ -0,0 +1,580 @@
...
+    G_OBJECT_CLASS (endless_shell_fx_wobbly_parent_class)->finalize (object);
+}
+
+static void
+endless_shell_fx_wobbly_init (EndlessShellFXWobbly *effect)
+{
+    EndlessShellFXWobblyPrivate *priv =
+        reinterpret_cast <EndlessShellFXWobblyPrivate *> (endless_shell_fx_wobbly_get_instance_private (effect));
+
+    priv->actor = nullptr;
```
cosimoc: I don't think you need to explicitly initialize these fields to 0/null.

src/Makefile.am
smspillaz: > given that upstream now does just that -- https://git.gnome.org/browse/gnome-shell/tree/configure.ac#n225

Looks like upstream is making `-Werror` a user option in that case.

Given that I don't think this part of the code is going to change much, I suppose we can remove the compiler flags. We gain consistency and aren't likely to pay any cost on maintainability.

js/ui/windowManager.js
smspillaz: I'm not a huge fan of `global`, but I suppose it isn't going to go away. Changed.

smspillaz:
All other comments addressed.

js/ui/windowManager.js
```diff
@@ -135,6 +137,67 @@ const DisplayChangeDialog = new Lang.Class({
     },
 });
 
+const EOSShellWobbly = new Lang.Class({
+    Name: 'EOSShellWobbly',
+    Extends: EndlessShellFX.Wobbly,
+    Properties: {
+        'settings' : GObject.ParamSpec.object('settings',
```
cosimoc: This property can be removed

src/wobbly-effect.cpp
```diff
@@ -0,0 +1,570 @@
+/*
+ * wobbly-effect.cpp
+ *
+ * Copyright (c) Endless Mobile Inc.
```
cosimoc: eos-desktop is GPLv2 -- add header

src/wobbly-effect.cpp
```diff
@@ -0,0 +1,570 @@
...
+        priv->model = nullptr;
+    }
+
+    if (priv->timeout_id)
+    {
+        g_source_remove (priv->timeout_id);
+        priv->timeout_id = 0;
+    }
+
+    if (priv->actor)
```
cosimoc: You could avoid storing the reference to the actor here entirely, since ClutterActorMeta will already keep a reference as long as the actor is alive.
In this case, you'd access the old reference before chaining up to parent in this function.

PR #162, opened 2016-09-02
T12729

smspillaz:
http://phabricator.endlessm.com/T12729

### DIFF
--- a/js/misc/missionGameService.js
+++ b/js/misc/missionGameService.js
@@ -49,8 +49,7 @@ const MissionChatboxTextService = new Lang.Class({
          * currently active task.
          *
          * The state of this service moves through each each task until
-         * it reaches the end, at which point it displays a message
-         * and sets itself to a 'done' state.
+         * it reaches the end, at which point these are set back to null.
          */
         this._openedForTheFirstTime = false;
         this._introLesson = null;
@@ -62,7 +61,13 @@ const MissionChatboxTextService = new Lang.Class({
         /* Connect to the service and refresh the content once we have a connection */
         Showmehow.ServiceProxy.new_for_bus(Gio.BusType.SESSION, 0, name, path, null,
                                            Lang.bind(this, function(source, result) {
-            this._service = Showmehow.ServiceProxy.new_for_bus_finish(result);
+            try {
+                this._service = Showmehow.ServiceProxy.new_for_bus_finish(result);
+            } catch (e) {
+                log('Error occurred in creating ShowmehowServiceProxy ' + String(e));
+                return;
+            }
+
             this._service.connect('lessons-changed', Lang.bind(this, function() {
                 /* When the underlying lessons change, we need to reset the entire
                  * state of this component and start from the beginning, including
@@ -71,10 +76,10 @@ const MissionChatboxTextService = new Lang.Class({
                  * Get the intro lesson again then reset all the state back
                  * to its initial point.
                  */
-                this._refreshContent(function() {
+                this._refreshContent(Lang.bind(this, function() {
                     this._openedForTheFirstTime = false;
                     this._currentTask = null;
-                });
+                }));
             }));
             this._service.connect('listening-for-lesson-events', Lang.bind(this, function(proxy, interestingEvents) {
                 this.emit('listening-for-events', interestingEvents.deep_unpack());
@@ -83,7 +88,14 @@ const MissionChatboxTextService = new Lang.Class({
         }));
     },
     _handleLessonResponse: function(source, result) {
-        let [success, returnValue] = this._service.call_attempt_lesson_remote_finish(result);
+        let success, returnValue;
+
+        try {
+            [success, returnValue] = this._service.call_attempt_lesson_remote_finish(result);
+        } catch (e) {
+            log('Error occurred in attempting lesson ' + String(e));
+            return;
+        }
 
         if (success) {
             let [responsesJSON, moveTo] = returnValue.deep_unpack();
@@ -174,7 +186,15 @@ const MissionChatboxTextService = new Lang.Class({
         };
 
         this._service.call_get_unlocked_lessons('console', null, Lang.bind(this, function(source, result) {
-            let [success, lessons] = this._service.call_get_unlocked_lessons_finish(result);
+            let success, lessons;
+
+            try {
+                [success, lessons] = this._service.call_get_unlocked_lessons_finish(result);
+            } catch (e) {
+                log('Error occurred in getting unlocked lessons ' + String(e));
+                return;
+            }
+
             lessons = lessons.deep_unpack();
 
             if (success) {
@@ -191,7 +211,15 @@ const MissionChatboxTextService = new Lang.Class({
         }));
 
         this._service.call_get_known_spells('console', null, Lang.bind(this, function(source, result) {
-            let [success, lessons] = this._service.call_get_known_spells_finish(result);
+            let success, lessons;
+
+            try {
+                [success, lessons] = this._service.call_get_known_spells_finish(result);
+            } catch (e) {
+                log('Error occurred in getting known spells ' + String(e));
+                return;
+            }
+
             lessons = lessons.deep_unpack();
 
             if (success) {
@@ -208,7 +236,14 @@ const MissionChatboxTextService = new Lang.Class({
         }));
 
         this._service.call_get_unlocked_lessons('shell', null, Lang.bind(this, function(source, result) {
-            let [success, lessons] = this._service.call_get_unlocked_lessons_finish(result);
+            let success, lessons;
+
+            try {
+                [success, lessons] = this._service.call_get_unlocked_lessons_finish(result);
+            } catch (e) {
+                log('Error occurred in getting unlocked lessons for shell ' + String(e));
+                return;
+            }
 
             if (success) {
                 /* There should be a single lesson here called introduction here. Save
@@ -236,7 +271,14 @@ const MissionChatboxTextService = new Lang.Class({
         }));
 
         this._service.call_get_clues('shell', null, Lang.bind(this, function(source, result) {
-            let [success, clues] = this._service.call_get_clues_finish(result);
+            let success, clues;
+
+            try {
+                [success, clues] = this._service.call_get_clues_finish(result);
+            } catch (e) {
+                log('Error occurred in getting clues ' + String(e));
+                return;
+            }
 
             if (success) {
                 this.emit('discover-new-inventory-items', clues.deep_unpack().map(function(clue) {
@@ -259,7 +301,15 @@ const MissionChatboxTextService = new Lang.Class({
 
         this._service.call_get_task_description('intro', taskName, null,
                                                 Lang.bind(this, function(source, result) {
-            let [success, returnValue] = this._service.call_get_task_description_finish(result);
+            let success, returnValue;
+
+            try {
+                [success, returnValue] = this._service.call_get_task_description_finish(result);
+            } catch (e) {
+                log(['Error occurred in getting task description for ',
+                     taskName, ' ', String(e)]);
+                return;
+            }
 
             if (success) {
                 let [desc, inputSpecString] = returnValue.deep_unpack();
--- a/js/misc/missionGameService.js
+++ b/js/misc/missionGameService.js
@@ -63,7 +63,7 @@ const MissionChatboxTextService = new Lang.Class({
             try {
                 this._service = Showmehow.ServiceProxy.new_for_bus_finish(result);
             } catch (e) {
-                log('Error occurred in creating ShowmehowServiceProxy ' + String(e));
+                logError(e, "Error occurred in creating ShowmehowServiceProxy");
                 return;
             }
 
@@ -92,7 +92,7 @@ const MissionChatboxTextService = new Lang.Class({
         try {
             [success, returnValue] = this._service.call_attempt_lesson_remote_finish(result);
         } catch (e) {
-            log('Error occurred in attempting lesson ' + String(e));
+            logError(e, 'Error occurred in attempting lesson');
             return;
         }
 
@@ -190,7 +190,7 @@ const MissionChatboxTextService = new Lang.Class({
             try {
                 [success, lessons] = this._service.call_get_unlocked_lessons_finish(result);
             } catch (e) {
-                log('Error occurred in getting unlocked lessons ' + String(e));
+                logError(e, 'Error occurred in getting unlocked lessons');
                 return;
             }
 
@@ -215,7 +215,12 @@ const MissionChatboxTextService = new Lang.Class({
             try {
                 [success, lessons] = this._service.call_get_known_spells_finish(result);
             } catch (e) {
-                log('Error occurred in getting known spells ' + String(e));
+                logError(e, 'Error occurred in getting known spells');
+                return;
+            }
+
+            if (!success) {
+                log('Warning: call to showmhow get_known_spells failed');
                 return;
             }
 
@@ -240,7 +245,7 @@ const MissionChatboxTextService = new Lang.Class({
             try {
                 [success, lessons] = this._service.call_get_unlocked_lessons_finish(result);
             } catch (e) {
-                log('Error occurred in getting unlocked lessons for shell ' + String(e));
+                logError(e, 'Error occurred in getting unlocked lessons for shell');
                 return;
             }
 
@@ -275,7 +280,7 @@ const MissionChatboxTextService = new Lang.Class({
             try {
                 [success, clues] = this._service.call_get_clues_finish(result);
             } catch (e) {
-                log('Error occurred in getting clues ' + String(e));
+                logError(e, 'Error occurred in getting clues');
                 return;
             }
 
@@ -305,8 +310,7 @@ const MissionChatboxTextService = new Lang.Class({
             try {
                 [success, returnValue] = this._service.call_get_task_description_finish(result);
             } catch (e) {
-                log(['Error occurred in getting task description for ',
-                     taskName, ' ', String(e)]);
+                logError(e, 'Error occurred in getting task description for ' + taskName);
                 return;
             }
 
--- a/js/misc/missionGameService.js
+++ b/js/misc/missionGameService.js
@@ -44,26 +44,26 @@ const MissionChatboxTextService = new Lang.Class({
         this._service = null;
 
         /* These properties make up the lesson's content. _lessonInfo refers
-         * to all of the text we need to display about the broader "lesson"
+         * to all of the text we need to display about the broader 'lesson'
          * and _taskInfo is all the text that we need to display about the
          * currently active task.
          *
          * The state of this service moves through each each task until
          * it reaches the end, at which point it displays a message
-         * and sets itself to a "done" state.
+         * and sets itself to a 'done' state.
          */
         this._openedForTheFirstTime = false;
         this._introLesson = null;
         this._currentTask = null;
 
-        let name = "com.endlessm.Showmehow.Service";
-        let path = "/com/endlessm/Showmehow/Service";
+        let name = 'com.endlessm.Showmehow.Service';
+        let path = '/com/endlessm/Showmehow/Service';
 
         /* Connect to the service and refresh the content once we have a connection */
         Showmehow.ServiceProxy.new_for_bus(Gio.BusType.SESSION, 0, name, path, null,
                                            Lang.bind(this, function(source, result) {
             this._service = Showmehow.ServiceProxy.new_for_bus_finish(result);
-            this._service.connect("lessons-changed", Lang.bind(this, function() {
+            this._service.connect('lessons-changed', Lang.bind(this, function() {
                 /* When the underlying lessons change, we need to reset the entire
                  * state of this component and start from the beginning, including
                  * showing any warnings.
@@ -76,8 +76,8 @@ const MissionChatboxTextService = new Lang.Class({
                     this._currentTask = null;
                 });
             }));
-            this._service.connect("listening-for-lesson-events", Lang.bind(this, function(proxy, interestingEvents) {
-                this.emit("listening-for-events", interestingEvents.deep_unpack());
+            this._service.connect('listening-for-lesson-events', Lang.bind(this, function(proxy, interestingEvents) {
+                this.emit('listening-for-events', interestingEvents.deep_unpack());
             }));
             this._refreshContent();
         }));
@@ -90,7 +90,7 @@ const MissionChatboxTextService = new Lang.Class({
             let responses = JSON.parse(responsesJSON);
 
             responses.forEach(Lang.bind(this, function(response) {
-                this.emit("chat-message", {
+                this.emit('chat-message', {
                     kind: response.type,
                     text: response.value
                 });
@@ -112,7 +112,7 @@ const MissionChatboxTextService = new Lang.Class({
                 }
             }
         } else {
-            log("Failed to call call_attempt_lesson_remote");
+            log('Failed to call call_attempt_lesson_remote');
         }
     },
     ready: function() {
@@ -132,14 +132,14 @@ const MissionChatboxTextService = new Lang.Class({
                     returnValue.deep_unpack().map(function(w) {
                         return w[0];
                     }).forEach(Lang.bind(this, function(w) {
-                        this.emit("chat-message", {
-                            kind: "scrolled",
-                            mode: "immediate",
+                        this.emit('chat-message', {
+                            kind: 'scrolled',
+                            mode: 'immediate',
                             text: w
                         });
                     }));
                 } else {
-                    log("Call to get_warnings_finish failed");
+                    log('Call to get_warnings_finish failed');
                 }
 
                 this._showTaskDescriptionForLesson(this._introLesson.entry);
@@ -157,68 +157,68 @@ const MissionChatboxTextService = new Lang.Class({
     },
     evaluate: function(text) {
         if (this._introLesson && this._currentTask) {
-            this._service.call_attempt_lesson_remote("intro", this._currentTask.name, text, null,
+            this._service.call_attempt_lesson_remote('intro', this._currentTask.name, text, null,
                                                      Lang.bind(this, this._handleLessonResponse));
         }
     },
     _refreshContent: function(completedCallback) {
         if (!this._service) {
-            log("Attempted to refresh content without a service!");
+            log('Attempted to refresh content without a service!');
         }
 
         var completion = {
-            "adventures": false,
-            "spells": false,
-            "inventory": false,
-            "intro": false
+            'adventures': false,
+            'spells': false,
+            'inventory': false,
+            'intro': false
         };
 
-        this._service.call_get_unlocked_lessons("console", null, Lang.bind(this, function(source, result) {
+        this._service.call_get_unlocked_lessons('console', null, Lang.bind(this, function(source, result) {
             let [success, lessons] = this._service.call_get_unlocked_lessons_finish(result);
             lessons = lessons.deep_unpack();
 
             if (success) {
-                this.emit("discover-new-adventures", lessons.map(function(lesson_spec) {
+                this.emit('discover-new-adventures', lessons.map(function(lesson_spec) {
                     return {
                         name: lesson_spec[0],
                         desc: lesson_spec[1]
                     };
                 }));
-                callWhenComplete(completion, "adventures", completedCallback);
+                callWhenComplete(completion, 'adventures', completedCallback);
             } else {
-                log("Warning: Call to showmehow get_unlocked_lessons failed");
+                log('Warning: Call to showmehow get_unlocked_lessons failed');
             }
         }));
 
-        this._service.call_get_known_spells("console", null, Lang.bind(this, function(source, result) {
+        this._service.call_get_known_spells('console', null, Lang.bind(this, function(source, result) {
             let [success, lessons] = this._service.call_get_known_spells_finish(result);
             lessons = lessons.deep_unpack();
 
             if (success) {
-                this.emit("discover-new-spells", lessons.map(function(lesson_spec) {
+                this.emit('discover-new-spells', lessons.map(function(lesson_spec) {
                     return {
                         name: lesson_spec[0],
                         desc: lesson_spec[1]
                     };
                 }));
-                callWhenComplete(completion, "spells", completedCallback);
+                callWhenComplete(completion, 'spells', completedCallback);
             } else {
-                log("Warning: Call to showmehow get_known_spells failed");
+                log('Warning: Call to showmehow get_known_spells failed');
             }
         }));
 
-        this._service.call_get_unlocked_lessons("shell", null, Lang.bind(this, function(source, result) {
+        this._service.call_get_unlocked_lessons('shell', null, Lang.bind(this, function(source, result) {
             let [success, lessons] = this._service.call_get_unlocked_lessons_finish(result);
 
             if (success) {
                 /* There should be a single lesson here called introduction here. Save
                  * it. */
                 lessons = lessons.deep_unpack().filter(function(lesson) {
-                    return lesson[0] == "intro";
+                    return lesson[0] == 'intro';
                 });
 
                 if (lessons.length !== 1) {
-                    log("Expected a single lesson for shell, cannot show intro lesson!");
+                    log('Expected a single lesson for shell, cannot show intro lesson!');
                     return;
                 }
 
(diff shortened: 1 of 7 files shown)


js/ui/status/missionGame.js
```diff
@@ -0,0 +1,71 @@
...
+        this.menu.addMenuItem(this._spells);
+
+        let name = "com.endlessm.Showmehow.Service";
+        let path = "/com/endlessm/Showmehow/Service";
+
+        Showmehow.ServiceProxy.new_for_bus(Gio.BusType.SESSION, 0, name, path, null,
+                                           Lang.bind(this, function(source, result) {
+            this._service = Showmehow.ServiceProxy.new_for_bus_finish(result);
+            this._service.call_get_unlocked_lessons(null, Lang.bind(this, function(source, result) {
+                [success, lessons] = this._service.call_get_unlocked_lessons_finish(result);
```
aperezdc: This is defining global variables, please prefix the line with `let` to use locals.

js/ui/status/missionGame.js
```diff
@@ -57,13 +57,22 @@ const Indicator = new Lang.Class({
                         this._adventures.menu.addAction(lesson_spec[1],
                                                         launchLessonAction(lesson_spec[0]));
                     }));
+                } else {
+                    log("Warning: Call to showmehow get_unlocked_lessons failed");
+                }
+            }));
 
+            this._service.call_get_known_spells(null, Lang.bind(this, function(source, result) {
+                [success, lessons] = this._service.call_get_known_spells_finish(result);
```
aperezdc: Prefix with `let` to avoid defining globals.

js/ui/status/missionGame.js
```diff
@@ -171,6 +171,28 @@ const ScrollingLabel = new Lang.Class({
...
+        this._text = params.text;
+        this.complete = false;
+    },
+    start: function(scrollView) {
+        this.complete = true;
+        this.set_text(this._text);
+        this.emit('finished-scrolling');
+        scrollView();
+    },
+    fastForward: function() {
```
aperezdc: Is this empty function used somewhere?

js/ui/status/missionGame.js
```diff
@@ -311,6 +318,26 @@ const Indicator = new Lang.Class({
                     log("Warning: Call to showmehow get_known_spells failed");
                 }
             }));
+
+            this._service.call_get_unlocked_lessons("shell", null, Lang.bind(this, function(source, result) {
+                [success, lessons] = this._service.call_get_unlocked_lessons_finish(result);
```
aperezdc: Please prefix with `let` to avoid defining globals.

js/ui/status/missionGame.js
```diff
@@ -323,19 +350,105 @@ const Indicator = new Lang.Class({
...
+                this._currentTaskText = null;
+                this._chatboxLessonCounter = 0;
+            }
+
+            return;
+        }
+
+        this._service.call_get_task_description("intro", lessonIndex, null,
+                                                Lang.bind(this, function(source, result) {
+            [success, returnValue] = this._service.call_get_task_description_finish(result);
```
aperezdc: Same: prefix this line with `let`.

js/ui/status/missionGame.js
```diff
@@ -392,7 +392,28 @@ const Indicator = new Lang.Class({
         this.menu.connect('open-state-changed', Lang.bind(this, function(menu, state) {
             if (state && !this._openedForTheFirstTime && this._introLesson) {
                 this._openedForTheFirstTime = true;
-                this._showTaskDescriptionForLesson(this._chatboxLessonCounter);
+
+                /* Get warnings and show them first, then show the first
+                 * chatbox description */
+                this._service.call_get_warnings(null, Lang.bind(this, function(source, result) {
+                    [success, returnValue] = this._service.call_get_warnings_finish(result);
```
aperezdc: Prefix with `let` to avoid creating globals.

js/ui/status/missionGame.js
```diff
@@ -387,6 +387,37 @@ const Indicator = new Lang.Class({
...
+            this._service.connect("lessons-changed", Lang.bind(this, function(proxy) {
+                /* When the underlying lessons change, we need to reset the entire
+                 * state of this component and start from the beginning, including
+                 * showing any warnings.
+                 *
+                 * Get the intro lesson again then reset all the state back
+                 * to its initial point.
+                 */
+                this._service.call_get_unlocked_lessons("shell", null, Lang.bind(this, function(source, result) {
+                    [success, lessons] = this._service.call_get_unlocked_lessons_finish(result);
```
aperezdc: Prefix with `let`.

js/ui/status/missionGame.js
```diff
@@ -260,6 +260,265 @@ function _addLabelToChatboxArea(label, chatboxResultsArea) {
...
+         * That depends on us having not been opened for the first
+         * time and the content being loaded. */
+        if (!this._openedForTheFirstTime && this._introLesson) {
+            this._openedForTheFirstTime = true;
+            this._chatboxLessonCounter = 0;
+
+            /* Get warnings and show them first, then show the first
+             * chatbox description */
+            this._service.call_get_warnings(null, Lang.bind(this, function(source, result) {
+                [success, returnValue] = this._service.call_get_warnings_finish(result);
```
aperezdc: Prefix with `let`, please.

js/misc/missionGameService.js
```diff
@@ -214,6 +215,23 @@ const MissionChatboxTextService = new Lang.Class({
                 log("Warning: Call to showmehow get_unlocked_lessons failed for intro lesson");
             }
         }));
+
+        this._service.call_get_clues("shell", null, Lang.bind(this, function(source, result) {
+            [success, clues] = this._service.call_get_clues_finish(result);
```
aperezdc: Add `let` at the start of the line.

js/misc/missionGameService.js
```diff
@@ -29,7 +29,7 @@ function callWhenComplete(completion, key, callback) {
     });
 
     if (complete && callback) {
-        callback(key);
+        callback(key. arguments);
```
aperezdc: Please remove the space between `key.` and `arguments`.

aperezdc:
Long PR, long review. Just commented with some nits. Other than that, all seems to be okay.

smspillaz:
@aperezdc Thanks for the review!

WRT to missing `let` statements, it might be good if one day we can start running `eslint` on `els-desktop`. That usually catches these type of things so that reviewers don't have to.

js/misc/missionGameService.js
smspillaz: Actually, this was a typo -> `.` should be `,` . Nonetheless, I'm not sure if `arguments` works that way now that I think of it.  I think you need to do `callback.apply(this, [key].concat(Array.from(arguments)));`

smspillaz:
@aperezdc Comments should be addressed in 37171d7d9 and 37171d7d9^

js/misc/missionGameService.js
```diff
@@ -0,0 +1,267 @@
+// -*- mode: js; js-indent-level: 4; indent-tabs-mode: nil -*-
+
+const Lang = imports.lang;
```
cosimoc: Nitpick: alphabetical order

js/misc/missionGameService.js
```diff
@@ -0,0 +1,267 @@
...
+         *
+         * The state of this service moves through each each task until
+         * it reaches the end, at which point it displays a message
+         * and sets itself to a "done" state.
+         */
+        this._openedForTheFirstTime = false;
+        this._introLesson = null;
+        this._currentTask = null;
+
+        const name = "com.endlessm.Showmehow.Service";
```
cosimoc: We use single quotes in the shell JS, except for translatable strings (multiple occurrences)

js/misc/missionGameService.js
```diff
@@ -0,0 +1,267 @@
...
+    Name: 'MissionChatboxTextService',
+    Extends: GLib.GObject,
+
+    _init: function(props) {
+        this.parent(props);
+
+        /* Null-initialise service for now, but we'll set it later */
+        this._service = null;
+
+        /* These properties make up the lesson's content. _lessonInfo refers
```
cosimoc: This comment mentions `._lessonInfo` and `._taskInfo` but they don't exist

js/misc/missionGameService.js
```diff
@@ -0,0 +1,267 @@
...
+ * Takes an object with keys mapping to bool values, a callback
+ * and a key name. Flips the key name to true and checks if
+ * they are all true, and if so, calls the callback.
+ *
+ * This is used by MissionChatboxTextService to ensure that
+ * all asynchronous tasks are completed before calling
+ * the completion callback. The callback is called with
+ * the key that triggered the callback.
+ */
+function callWhenComplete(completion, key, callback) {
```
cosimoc: There is code with a similar concept at https://github.com/endlessm/eos-desktop/blob/master/js/gdm/batch.js -- would be nice if we could reuse that

js/misc/missionGameService.js
```diff
@@ -0,0 +1,267 @@
...
+        this._currentTask = null;
+
+        const name = "com.endlessm.Showmehow.Service";
+        const path = "/com/endlessm/Showmehow/Service";
+
+        /* Connect to the service and refresh the content once we have a connection */
+        Showmehow.ServiceProxy.new_for_bus(Gio.BusType.SESSION, 0, name, path, null,
+                                           Lang.bind(this, function(source, result) {
+            this._service = Showmehow.ServiceProxy.new_for_bus_finish(result);
+            this._service.connect("lessons-changed", Lang.bind(this, function(proxy) {
```
cosimoc: `proxy` is unused and could be omitted

js/misc/missionGameService.js
```diff
@@ -0,0 +1,267 @@
...
+                /* When the underlying lessons change, we need to reset the entire
+                 * state of this component and start from the beginning, including
+                 * showing any warnings.
+                 *
+                 * Get the intro lesson again then reset all the state back
+                 * to its initial point.
+                 */
+                this._refreshContent(function() {
+                    this._openedForTheFirstTime = false;
```
cosimoc: You'll have to bind the callback to the `this` instance for these to be effective

js/misc/missionGameService.js
```diff
@@ -0,0 +1,267 @@
...
+            }));
+            this._refreshContent();
+        }));
+    },
+    commenceIntroLesson: function() {
+        /* If possible in the current state, commence the intro lesson.
+         *
+         * That depends on us having not been opened for the first
+         * time and the content being loaded. */
+        if (!this._openedForTheFirstTime && this._introLesson) {
```
cosimoc: Prefer early returns

js/misc/missionGameService.js
```diff
@@ -0,0 +1,267 @@
...
+        this._introLesson = null;
+        this._currentTask = null;
+
+        const name = "com.endlessm.Showmehow.Service";
+        const path = "/com/endlessm/Showmehow/Service";
+
+        /* Connect to the service and refresh the content once we have a connection */
+        Showmehow.ServiceProxy.new_for_bus(Gio.BusType.SESSION, 0, name, path, null,
+                                           Lang.bind(this, function(source, result) {
+            this._service = Showmehow.ServiceProxy.new_for_bus_finish(result);
```
cosimoc: Shouldn't this be in a try/catch?

js/misc/missionGameService.js
```diff
@@ -0,0 +1,267 @@
...
+                 * to its initial point.
+                 */
+                this._refreshContent(function() {
+                    this._openedForTheFirstTime = false;
+                    this._currentTask = null;
+                });
+            }));
+            this._refreshContent();
+        }));
+    },
```
cosimoc: Add empty line between methods

js/misc/missionGameService.js
```diff
@@ -0,0 +1,267 @@
...
+         *
+         * That depends on us having not been opened for the first
+         * time and the content being loaded. */
+        if (!this._openedForTheFirstTime && this._introLesson) {
+            this._openedForTheFirstTime = true;
+
+            /* Get warnings and show them first, then show the first
+             * chatbox description */
+            this._service.call_get_warnings(null, Lang.bind(this, function(source, result) {
+                const [success, returnValue] = this._service.call_get_warnings_finish(result);
```
cosimoc: This (and other `finish` calls) should be in try/catch blocks

js/misc/missionGameService.js
```diff
@@ -0,0 +1,267 @@
...
+                } else {
+                    log("Call to get_warnings_finish failed");
+                }
+
+                this._showTaskDescriptionForLesson(this._introLesson.entry);
+            }));
+        }
+    },
+    evaluate: function(text) {
+        if (this._introLesson && this._currentTask) {
```
cosimoc: Prefer early returns

js/ui/panel.js
```diff
@@ -40,6 +40,8 @@ const ICON_ENTER_ANIMATION_DELTA = PANEL_ICON_SIZE * 0.25;
 const ICON_ENTER_ANIMATION_SPEED = ICON_ENTER_ANIMATION_DELTA * 0.0012;
 const ICON_ENTER_ANIMATION_DELAY = ICON_ENTER_ANIMATION_SPEED * ICON_ENTER_ANIMATION_DELTA * 0.5 + PANEL_ANIMATION_TIME;
 
+const SHELL_SETTINGS = new Gio.Settings({ schema: 'org.gnome.shell' });
```
cosimoc: You can use `global.settings`

js/ui/panel.js
```diff
@@ -333,6 +335,14 @@ try {
...
 }
 
+if (SHELL_SETTINGS.get_boolean('enable-mission-game')) {
+    PANEL_ITEM_IMPLEMENTATIONS['missionGame'] = imports.ui.status.missionGame.MissionGameIndicator;
+    PANEL_ITEM_IMPLEMENTATIONS['missionGameToolbox'] = imports.ui.status.missionGame.MissionGameToolboxIndicator;
+    PANEL_ITEM_IMPLEMENTATIONS['missionGameChatbox'] = imports.ui.status.missionGame.MissionGameChatboxIndicator;
+    log('Mission game enabled');
+}
+
+
```
cosimoc: Remove empty line

js/ui/status/missionGame.js
```diff
@@ -0,0 +1,528 @@
...
+ * the parent and hope for the best.
+ */
+function createSubMenuMenuItemWithFauxParent(name, parent) {
+    let item = new PopupMenu.PopupSubMenuMenuItem(name);
+    item.menu._getTopMenu = Lang.bind(item, function() {
+        return parent;
+    });
+    return item;
+}
+
```
cosimoc: Remove empty line

js/ui/status/missionGame.js
```diff
@@ -0,0 +1,528 @@
...
+    Lang.copyProperties(src, dst);
+    return dst;
+}
+
+const ScrolledLabel = new Lang.Class({
+    Name: 'ScrolledLabel',
+    Extends: St.Label,
+
+    _init: function(params) {
+        let parentParams = copyObject(params);
```
cosimoc: This way of using parameters by copying is very non-idiomatic. Why do you need to copy them?

js/ui/status/missionGame.js
```diff
@@ -0,0 +1,528 @@
...
+        this._text = params.text;
+        this._textIndex = 0;
+        this._scrollTimer = 0;
+        this._waitCount = 0;
+        this.complete = false;
+
+        this.style_class = 'chatbox-character-text';
+
+        this.set_x_expand(true);
+        this.clutter_text.set_x_expand(true);
```
cosimoc: Nitpick: it's usually preferred to set properties for simple values like these

js/ui/status/missionGame.js
```diff
@@ -0,0 +1,528 @@
...
+        parentParams.text = '';
+
+        this.parent(parentParams);
+        this._text = params.text;
+        this._textIndex = 0;
+        this._scrollTimer = 0;
+        this._waitCount = 0;
+        this.complete = false;
+
+        this.style_class = 'chatbox-character-text';
```
cosimoc: You could add this to the parameters you pass to `this.parent()`

js/ui/status/missionGame.js
cosimoc: (Usually `Params.parse()` or something similar is used instead)

js/ui/status/missionGame.js
```diff
@@ -0,0 +1,528 @@
...
+    }
+});
+Signals.addSignalMethods(ScrolledLabel.prototype);
+
+WrappedLabel = new Lang.Class({
+    Name: 'WrappedLabel',
+    Extends: ScrolledLabel,
+    _init: function(params) {
+        this.parent(params);
+        this._text = wrapTextWith(this._text, WRAP_CONSTANT, "> ").join("\n");
```
cosimoc: You can pass the text in the parameters instead of setting a private field from outside like this.

js/ui/status/missionGame.js
```diff
@@ -0,0 +1,528 @@
...
+
+        this.style_class = 'chatbox-user-text';
+
+        this.set_x_expand(true);
+        this.clutter_text.set_x_expand(true);
+    },
+    start: function(scrollView) {
+        this.set_text(this._text);
+        this["translation-y"] = 5;
+        this["opacity"] = 0;
```
cosimoc: Prefer the usual `this._translation_y` etc notation

js/ui/status/missionGame.js
```diff
@@ -0,0 +1,528 @@
...
+
+        this.parent(parentParams);
+
+        /* Retain service, which evaluates text entry */
+        this._service = service;
+
+        /* Setup layout and style */
+        const margin = 10;
+        ["top", "bottom", "left", "right"].forEach(Lang.bind(this, function(d) {
+            this["margin-" + d] = margin
```
cosimoc: Missing semi

js/ui/status/missionGame.js
```diff
@@ -0,0 +1,528 @@
...
+        }));
+
+        this.set_size(400, 450);
+        this.style_class = 'chatbox-text-container';
```
cosimoc: You could pass all of these to `this.parent()`

js/ui/status/missionGame.js
```diff
@@ -0,0 +1,528 @@
...
+
+
+/* Combined indicator option */
+const MissionGameIndicator = new Lang.Class({
+    Name: 'MissionGameIndicator',
+    Extends: PanelMenu.SystemStatusButton,
+
+    _init: function() {
+        this.parent('folder-drag-accept-symbolic', _("Mission"));
+        this.setIcon('folder-drag-accept-symbolic');
```
cosimoc: Is chaining up with the icon name not enough?

js/ui/status/missionGame.js
```diff
@@ -0,0 +1,528 @@
...
+
+        this._service = new MissionGameService.MissionChatboxTextService();
+
+        /* Create layout for indicator menu */
+        const hbox = new St.BoxLayout({name: 'menuArea'});
```
cosimoc: `const` is only used for constants or imports in the shell coding style (many occurrences). In this case, typically you'd use `let`

js/misc/missionGameService.js
```diff
@@ -0,0 +1,267 @@
...
+    });
+
+    if (complete && callback) {
+        callback.apply(this, [key].concat(Array.from(arguments)));
+    }
+}
+
+const MissionChatboxTextService = new Lang.Class({
+    Name: 'MissionChatboxTextService',
+    Extends: GLib.GObject,
```
cosimoc: There should be no need for this to be a GObject

js/ui/status/missionGame.js
```diff
@@ -0,0 +1,528 @@
...
+/* Combined indicator option */
+const MissionGameIndicator = new Lang.Class({
+    Name: 'MissionGameIndicator',
+    Extends: PanelMenu.SystemStatusButton,
+
+    _init: function() {
+        this.parent('folder-drag-accept-symbolic', _("Mission"));
+        this.setIcon('folder-drag-accept-symbolic');
+
+        this._service = new MissionGameService.MissionChatboxTextService();
```
cosimoc: In general, the service connection to should probably be a singleton

js/ui/status/missionGame.js
```diff
@@ -0,0 +1,528 @@
...
+            this._inventory.menu.removeAll();
+            items.forEach(Lang.bind(this, function(item) {
+                this._inventory.menu.addAction(item.name, function() {
+                });
+            }));
+        }));
+    }
+});
+
+function createSeparator() {
```
cosimoc: Can this be shared with the other separators in the shell?

js/ui/status/missionGame.js
smspillaz: Ah thanks, thats really helpful. copyObject was used here so that we dont pass params to base objects that they dont understand, but if Params.parse can help us deal with that then all the better.

js/ui/status/missionGame.js
smspillaz: It could be. I kept it here for the sake of not polluting the other parts of the shell codebase but otherwise I'd be in favour of abstracting it out.

js/misc/missionGameService.js
smspillaz: Now that this function generally updates state based on a few different conditions, we need the if-block here.

js/misc/missionGameService.js
smspillaz: It emits signals, which is why I had it as a GObject. Is there a way to have normal JSObjects emit signals without being bound to a GObject instance?

js/misc/missionGameService.js
smspillaz: Fixed in 1431aaa7cb7ed3dfad5f080bbe84f2655a9a5ccb

js/misc/missionGameService.js
smspillaz: Fixed in 5c35f5e2622633cfe2d29e3e240331964d8b4c2a

js/misc/missionGameService.js
smspillaz: Indeed - I'll make a note of it here and add it on the next round of review.

js/misc/missionGameService.js
smspillaz: Looks like this was fixed in 1431aaa7

js/misc/missionGameService.js
smspillaz: Fixed in 07b2e4d8

js/misc/missionGameService.js
smspillaz: Agreed that it is better to handle any exceptions here. Fixed in 07b2e4d8b97b962886fd82bdd350991a2a9a52fb

js/misc/missionGameService.js
smspillaz: Will fix on next review round.

js/misc/missionGameService.js
smspillaz: Fixed in 07b2e4d8b97b962886fd82bdd350991a2a9a52fb

js/ui/panel.js
smspillaz: Fixed in fd93e96fbd4a682b910b90c66fbc8246ae468395

js/ui/panel.js
smspillaz: Fixed in df5a428d7dcae8d61fe53edc19f7883507580a1b

js/ui/status/missionGame.js
```diff
@@ -0,0 +1,528 @@
+// -*- mode: js; js-indent-level: 4; indent-tabs-mode: nil -*-
+
+const Clutter = imports.gi.Clutter;
+const Lang = imports.lang;
```
smspillaz: Fixed in c0db11c269500f5f83fef2b4561e490f8d89b96a

js/ui/status/missionGame.js
smspillaz: Killed copyObject in17a0e6c7c2bc8650a65a7a9722256d749a56bc42

js/ui/status/missionGame.js
smspillaz: Agreed - 71e1cc1c5d98863031ed6cd0c02dc4f64927ecc3

js/ui/status/missionGame.js
smspillaz: Yes - db7058aa2c0e89b71249fd3ed97f1698e33f0fa6

js/ui/status/missionGame.js
smspillaz: Done - 47a2f4357ee13c957abbbabbcbef679824be9de1

js/ui/status/missionGame.js
smspillaz: This class was dropped in fe989841d2cba73205f1e051be9dbb5ee416bc17 so no longer necessary :)

js/ui/status/missionGame.js
smspillaz: Fixed

js/ui/status/missionGame.js
smspillaz: Done c5cd34fd00d4cd45329f2969a7eafc7284f47c10

js/ui/status/missionGame.js
smspillaz: I've given it a try in ea7395721b38d2893846fd15c19372ac0a4e10fe , will test now to see if that still works.

js/ui/status/missionGame.js
smspillaz: Clarified the policy here after our slack discussion last week. Fixed in fba888932eb6dd0250b01903bbbed56148d6b13d

js/ui/status/missionGame.js
smspillaz: Done - fbed314490478dc3c08c5c184336ca2063448a76

smspillaz:
Okay, most comments should be addressed. My apologies that the turnaround time on this was so long - this review came up right in the middle of a big refactor and then me getting sick / flooded with university work.

js/ui/status/missionGame.js
```diff
@@ -0,0 +1,648 @@
...
+         * it in a different style and add it to the chatbox */
+        this._service.connect('chat-message', Lang.bind(this, function(chat, message) {
+            /* First try and append it to the last label. If so, no work to do */
+            if (this._appendToLastLabelInChatboxResultsArea(message)) {
+                return;
+            }
+
+            /* Create a new bubble */
+            let classes = {
+                'scrolled': ScrolledLabel,
```
cosimoc: You should not need single quotes for these

js/ui/status/missionGame.js
```diff
@@ -0,0 +1,648 @@
...
+
+            if (message.mode === 'immediate') {
+                label.fastForward();
+            }
+
+            this._pushLabelToChatboxResultsArea(label);
+        }));
+        this._service.connect('user-input-bubble', Lang.bind(this, function(service, bubbleConfig, lesson, task) {
+            let classes = {
+                'text': TextResponseArea,
```
cosimoc: Also these should not use quotes

js/ui/status/missionGame.js
```diff
@@ -0,0 +1,648 @@
...
+ * Helper function to add a ScrollableLabel to a chatbox results area.
+ */
+function _addLabelToChatboxArea(label, chatboxResultsArea) {
+    if (label.view) {
+        chatboxResultsArea.add(label.view);
+    }
+
+    /* Start scrolling the label */
+    label.start(function() {
+        chatboxResultsArea.vadjustment.set_value(Number.MAX_VALUE);
```
cosimoc: Should this not be `vadjustment.upper`?

js/ui/status/missionGame.js
```diff
@@ -0,0 +1,648 @@
...
+    },
+    acceptAdditionalContent: function() {
+        return false;
+    }
+});
+Signals.addSignalMethods(ChoiceResponseArea.prototype);
+
+const ExternalEventsResponseArea = new Lang.Class({
+    Name: 'ExternalEventsResponseArea',
+    Extends: GLib.Object,
```
cosimoc: I don't think this is needed

js/ui/status/missionGame.js
```diff
@@ -0,0 +1,648 @@
...
+Signals.addSignalMethods(TextResponseArea.prototype);
+
+const ConsoleResponseArea = new Lang.Class({
+    Name: 'ConsoleResponseArea',
+    Extends: TextResponseAreaBase,
+    _init: function(params) {
+        this.parent(params, '$ ');
+    }
+});
+Signals.addSignalMethods(ConsoleResponseArea.prototype);
```
cosimoc: I think you can do this just in the base class

js/ui/status/missionGame.js
```diff
@@ -0,0 +1,648 @@
...
+const WrappedLabel = new Lang.Class({
+    Name: 'WrappedLabel',
+    Extends: ScrolledLabel,
+    _init: function(params, settings) {
+        this.parent(params, settings);
+        this._text = wrapTextWith(settings.text, WRAP_CONSTANT, '> ').join("\n");
+        this.fastForward();
+    },
+    acceptAdditionalContent: function(content) {
+        if (content.text && content.text.length && content.type === 'wrapped') {
```
cosimoc: I think `content.text` will evaluate to true if the length is not zero, so the second check is redundant

js/ui/status/missionGame.js
cosimoc: Still relevant

js/ui/status/missionGame.js
```diff
@@ -0,0 +1,648 @@
...
+const BoxPointer = imports.ui.boxpointer;
+const PanelMenu = imports.ui.panelMenu;
+const PopupMenu = imports.ui.popupMenu;
+
+const MissionGameService = imports.misc.missionGameService;
+
+function launchLessonAction(lesson) {
+    return function() {
+        /* XXX: This needs to spawn a wrapper script that goes through each lesson
+         * individually as opposed to just running showmehow. */
```
cosimoc: I feel that this should be somehow coordinated with the service over DBus, or something a bit more robust than this.

js/misc/missionGameService.js
smspillaz: Yes, that's correct. It won't do anything unless the content is loaded.

js/ui/status/missionGame.js
```diff
@@ -0,0 +1,648 @@
...
+        this._service.connect('discover-new-spells', Lang.bind(this, function(chat, spells) {
+            this._spells.menu.removeAll();
+            spells.forEach(Lang.bind(this, function(spell) {
+                this._spells.menu.addAction(spell.desc, launchLessonAction(spell.name));
+            }));
+        }));
+        this._service.connect('discover-new-inventory-items', Lang.bind(this, function(chat, items) {
+            this._inventory.menu.removeAll();
+            items.forEach(Lang.bind(this, function(item) {
+                this._inventory.menu.addAction(item.name, function() {
```
smspillaz: It looks like `addAction` calls `callback` unconditionally (`popupMenu:823`)

js/ui/status/missionGame.js
```diff
@@ -0,0 +1,648 @@
...
+        this.parent('starred-symbolic', _('Mission'));
+        this.setIcon('starred-symbolic');
+
+        this._service = MissionGameService.getService();
+
+        /* Create layout for indicator menu */
+        let hbox = new St.BoxLayout({name: 'menuArea'});
+
+        /* Add toolbox, separator, chatbox */
+        hbox.add(new MissionToolbox({}, this.menu, this._service));
```
smspillaz: They're options for testing based on @dalio48 's feedback. Option 1 is to have a toolbox and a chatbox indicator, and then option 2 is to have them both merged into one. I probably should have added a comment about that.

js/ui/windowManager.js
```diff
@@ -752,6 +754,26 @@ const WindowManager = new Lang.Class({
...
+
+        this._missionGameService = MissionGameService.getService();
+
+        /* For mission we add another event handler for grab-op-begin
+         * to see when a window has moved or is moving.
+         *
+         * This is expensive, so we only do it if we are interested
+         * in window-move events */
+        this._missionGameService.connect('listening-for-events', Lang.bind(this, function(service, events) {
+            if (events.indexOf('window-moved')) {
```
smspillaz: Yes, of course.

smspillaz:
Okay, most of the smaller follow-up comments should be addressed. I'll keep this in dev for now, since I still need to address some of the larger issues (use of GObject) tomorrow morning. This will be ready for another round of review by Tuesday.

js/misc/missionGameService.js
smspillaz: Ah, right, having a look at the context, I see your point. I'll return early. Thanks.

js/misc/missionGameService.js
smspillaz: Interesting. I tried that just now and it works - surprising that it does. Do we have a convention for how we define "normal" objects? You can probably see what I used in later commits, but I'm not sure if that's the convention.

smspillaz: Actually, having had a look at https://github.com/endlessm/eos-desktop/blob/master/js/gdm/batch.js - it seems like `Lang.Class` supports not providing an `Extends` (which defaults to `_Base`).

I much prefer the `Lang.Class` style to just calling constructors directly, so I think I'll revert back to using that and not put `GObject.Object` in the `Extends`

js/ui/status/missionGame.js
smspillaz: Hmm, I'm not sure what empty line you're referring to here.

smspillaz:
https://github.com/endlessm/eos-desktop/pull/162/commits/e41394e6d736afb3ad163e3b2b5672d8a3d8783b should be a candidate for upstreaming

smspillaz:
This branch can be dropped for now, given that we are looking at doing the same thing as a GTK+ app. However, the stuff to track shell interactions will need to be retained.

PR #163, opened 2016-09-02
T12765

smspillaz:
http://phabricator.endlessm.com/T12765

### DIFF
--- a/js/ui/windowManager.js
+++ b/js/ui/windowManager.js
@@ -585,30 +585,6 @@ const TilePreview = new Lang.Class({
     }
 });
 
-/**
- * Determine all windows belonging to a certain process id
- */
-function windowActorsFromPid(pid) {
-    return global.get_window_actors().filter(function(window_actor) {
-        return window_actor.get_meta_window().get_pid() == pid;
-    });
-}
-
-/**
- * For a given process id, determine the corresponding window
- * (if any) and its size.
- */
-function pidToActorInfo(pid) {
-    let windowActors = windowActorsFromPid(pid);
-
-    return {
-        window: windowActors.length ? windowActors[0] : null,
-        rect: (windowActors.length ?
-               windowActors[0].get_meta_window().get_frame_rect() : null),
-        pid: pid,
-    };
-}
-
 const WindowManager = new Lang.Class({
     Name: 'WindowManager',
 
@@ -779,6 +755,23 @@ const WindowManager = new Lang.Class({
     },
 
     _handleRotateBetweenPidWindows: function(proxy, sender, [src, dst]) {
+        /**
+         * For a given process id, determine the corresponding window
+         * (if any) and its size.
+         */
+        function pidToActorInfo(pid) {
+            let windowActors = global.get_window_actors().filter(function(windowActor) {
+                return windowActor.get_meta_window().get_pid() == pid;
+            });
+
+            return {
+                window: windowActors.length ? windowActors[0] : null,
+                rect: (windowActors.length ?
+                       windowActors[0].get_meta_window().get_frame_rect() : null),
+                pid: pid,
+            };
+        }
+
         let srcActorInfo = pidToActorInfo(src);
         this._pendingRotateAnimations.push({
             src: srcActorInfo,
--- a/js/ui/windowManager.js
+++ b/js/ui/windowManager.js
@@ -821,7 +821,7 @@ const WindowManager = new Lang.Class({
                 animationSpec.src.window.pivot_point = new Clutter.Point({ x: 0.5, y: 0.5 });
                 animationSpec.src.window.set_cull_back_face(true);
                 animationSpec.dst.window.set_cull_back_face(true);
-                animationSpec.dst.window['opacity'] = 0;
+                animationSpec.dst.window.opacity = 0;
                 let dst_geometry = animationSpec.src.rect;
                 animationSpec.dst.window.get_meta_window().move_resize_frame(false,
                                                                              dst_geometry.x,
--- a/js/ui/windowManager.js
+++ b/js/ui/windowManager.js
@@ -1370,7 +1370,7 @@ const WindowManager = new Lang.Class({
             Tweener.addTween(actor,
                              { scale_x: 1,
                                scale_y: 1,
-                               time: WINDOW_ANIMATION_TIME * 20, // Entire animation takes twice the normal time,
+                               time: WINDOW_ANIMATION_TIME * 2, // Entire animation takes twice the normal time,
                                                                 // but it appears to take about the same duration
                                transition: function(t, b, c, d) {
                                    // Easing function similar to easeOutElastic, but less aggressive.
--- a/js/ui/windowManager.js
+++ b/js/ui/windowManager.js
@@ -586,8 +586,8 @@ const TilePreview = new Lang.Class({
     }
 });
 
-const DisableBackfaceCullingEffect = new Lang.Class({
-    Name: 'DisableBackfaceCullingEffect',
+const EnableBackfaceCullingEffect = new Lang.Class({
+    Name: 'EnableBackfaceCullingEffect',
     Extends: Clutter.Effect,
     _init : function(params) {
         this.parent(params);
@@ -817,10 +817,14 @@ const WindowManager = new Lang.Class({
                                                                              dst_geometry.width,
                                                                              dst_geometry.height);
 
-                animationSpec.src.window.add_effect_with_name('disable-culling',
-                                                              new DisableBackfaceCullingEffect());
-                animationSpec.dst.window.add_effect_with_name('disable-culling',
-                                                              new DisableBackfaceCullingEffect());
+                animationSpec.src.window.add_effect_with_name('enable-culling',
+                                                              new EnableBackfaceCullingEffect());
+                animationSpec.dst.window.add_effect_with_name('enable-culling',
+                                                              new EnableBackfaceCullingEffect());
+
+                /* Tween both windows in a rotation animation at the same time
+                 * with backface culling enabled on both. This will allow for
+                 * a smooth transition. */
                 Tweener.addTween(animationSpec.src.window, {
                     "rotation-angle-y": -120,
                     time: WINDOW_ANIMATION_TIME,
--- a/js/ui/windowManager.js
+++ b/js/ui/windowManager.js
@@ -2,6 +2,7 @@
 
 const Cairo = imports.cairo;
 const Clutter = imports.gi.Clutter;
+const Cogl = imports.gi.Cogl;
 const Gdk = imports.gi.Gdk;
 const GLib = imports.gi.GLib;
 const Gio = imports.gi.Gio;
@@ -37,6 +38,20 @@ const SKYPE_WINDOW_CLOSE_TIMEOUT_MS = 1000;
 
 const DISPLAY_REVERT_TIMEOUT = 30; // in seconds - keep in sync with mutter
 const ONE_SECOND = 1000; // in ms
+
+const SylvesterServiceIface = '<node> \
+  <interface name="com.endlessm.Sylvester.Service"> \
+    <method name="DownloadSourcesForPID"> \
+      <arg name="pid" direction="in" type="i"/> \
+    </method> \
+    <signal name="RotateBetweenPidWindows"> \
+      <arg name="src" type="i"/> \
+      <arg name="dest" type="i"/> \
+    </signal> \
+  </interface> \
+</node>';
+
+const SylvesterService = Gio.DBusProxy.makeProxyWrapper(SylvesterServiceIface);
 
 const DisplayChangeDialog = new Lang.Class({
     Name: 'DisplayChangeDialog',
@@ -571,6 +586,20 @@ const TilePreview = new Lang.Class({
     }
 });
 
+const DisableBackfaceCullingEffect = new Lang.Class({
+    Name: 'DisableBackfaceCullingEffect',
+    Extends: Clutter.Effect,
+    _init : function(params) {
+        this.parent(params);
+    },
+    vfunc_pre_paint: function() {
+        Cogl.set_backface_culling_enabled(true);
+    },
+    vfunc_post_paint: function() {
+        Cogl.set_backface_culling_enabled(false);
+    }
+});
+
 const WindowManager = new Lang.Class({
     Name: 'WindowManager',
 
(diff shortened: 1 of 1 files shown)


po/eos-desktop.pot
```diff
@@ -1,1350 +0,0 @@
-# SOME DESCRIPTIVE TITLE.
```
aperezdc: It looks like this file is being removed by your changes accidentally. Please amend the last commit so the file is not deleted.

aperezdc:
Without knowing a lot about Clutter, seems to be all good — check the deleted `.pot` file, though.

js/ui/windowManager.js
```diff
@@ -717,6 +784,126 @@ const WindowManager = new Lang.Class({
...
+        this._sylvesterListener = new SylvesterService(Gio.DBus.session,
+                                                       'com.endlessm.Sylvester.Service',
+                                                       '/com/endlessm/Sylvester/Service');
+        this._sylvesterListener.connectSignal('RotateBetweenPidWindows',
+                                              Lang.bind(this, this._handleRotateBetweenPidWindows));
+        this._pendingRotateAnimations = [];
+        this._rotateOutActors = [];
+        this._rotateInActors = [];
+        this._rotateInTimeouts = [];
+        log("Connected to Sylvester");
```
cosimoc: Remove this

js/ui/windowManager.js
```diff
@@ -571,6 +586,44 @@ const TilePreview = new Lang.Class({
     }
 });
 
+const EnableBackfaceCullingEffect = new Lang.Class({
+    Name: 'EnableBackfaceCullingEffect',
+    Extends: Clutter.Effect,
+    _init : function(params) {
```
cosimoc: This can be removed entirely, since it just calls the parent

js/ui/windowManager.js
```diff
@@ -717,6 +784,126 @@ const WindowManager = new Lang.Class({
...
+                                              Lang.bind(this, this._handleRotateBetweenPidWindows));
+        this._pendingRotateAnimations = [];
+        this._rotateOutActors = [];
+        this._rotateInActors = [];
+        this._rotateInTimeouts = [];
+        log("Connected to Sylvester");
+    },
+
+    _handleRotateBetweenPidWindows: function(proxy, sender, [src, dst]) {
+        const srcActorInfo = pidToActorInfo(src)
```
cosimoc: Missing semi

js/ui/windowManager.js
```diff
@@ -571,6 +586,44 @@ const TilePreview = new Lang.Class({
...
+        return window_actor.get_meta_window().get_pid() == pid;
+    });
+}
+
+/**
+ * For a given process id, determine the corresponding window
+ * (if any) and its size.
+ */
+function pidToActorInfo(pid) {
+    const windowActors = windowActorsFromPid(pid);
```
cosimoc: Don't use `const` except for imports and actual constants (multiple occurrences)

js/ui/windowManager.js
```diff
@@ -717,6 +784,126 @@ const WindowManager = new Lang.Class({
...
+    _handleRotateBetweenPidWindows: function(proxy, sender, [src, dst]) {
+        const srcActorInfo = pidToActorInfo(src)
+        this._pendingRotateAnimations.push({
+            src: srcActorInfo,
+            dst: pidToActorInfo(dst)
+        });
+        this._updateReadyRotateAnimationsWith(srcActorInfo.actor);
+    },
+    
+    _updateReadyRotateAnimationsWith: function (window) {
```
cosimoc: Remove space before paren

js/ui/windowManager.js
```diff
@@ -717,6 +784,126 @@ const WindowManager = new Lang.Class({
...
+                 * texture will be defined. Until that point, the contents are just empty and
+                 * so we can't just rely on map events to get to where we want. */
+                animationSpec.dst.window.rotation_angle_y = -180;
+                animationSpec.dst.window.pivot_point = new Clutter.Point({ x: 0.5, y: 0.5 });
+                animationSpec.src.window.pivot_point = new Clutter.Point({ x: 0.5, y: 0.5 });
+                animationSpec.src.window.add_effect_with_name('enable-culling',
+                                                              new EnableBackfaceCullingEffect());
+                animationSpec.dst.window.add_effect_with_name('enable-culling',
+                                                              new EnableBackfaceCullingEffect());
+                animationSpec.dst.window['opacity'] = 0;
```
cosimoc: Just use `window.opacity`

js/ui/windowManager.js
```diff
@@ -571,6 +586,44 @@ const TilePreview = new Lang.Class({
...
+
+/**
+ * For a given process id, determine the corresponding window
+ * (if any) and its size.
+ */
+function pidToActorInfo(pid) {
+    const windowActors = windowActorsFromPid(pid);
+
+    return {
+        actor: windowActors.length ? windowActors[0] : null,
```
cosimoc: If I understand the code correctly, what you call here `actor`, you later refer to as `window`. So the code will effectively only end up working when none of the two apps are started.

js/ui/windowManager.js
```diff
@@ -609,6 +662,20 @@ const WindowManager = new Lang.Class({
             this._unminimizeWindowDone(shellwm, actor);
             this._mapWindowDone(shellwm, actor);
             this._destroyWindowDone(shellwm, actor);
+            this._rotateInCompleted(actor);
+            this._rotateOutCompleted(actor);
+
+            let activeWaitForRotateTimeout = this._rotateInTimeouts.filter(function(timeout) {
+                return timeout == actor._waitingRotateTimeout;
+            });
```
cosimoc: This looks unnecessarily complicated. You should always remove `actor._waitingRotateTimeout` if it's set.

js/ui/windowManager.js
```diff
@@ -609,6 +662,20 @@ const WindowManager = new Lang.Class({
...
+
+            if (activeWaitForRotateTimeout.length === 1) {
+                GLib.source_remove(activeWaitForRotateTimeout[0]);
+                this._rotateInTimeouts = this.rotateInTimeouts.filter(function(timeout) {
```
cosimoc: I think this could be made simpler by just using `indexOf()` and `splice()` to remove the element.

js/ui/windowManager.js
```diff
@@ -717,6 +784,126 @@ const WindowManager = new Lang.Class({
...
+                        unsatisfiedPids++;
+                    }
+                }
+            });
+            
+            if (unsatisfiedPids == 0) {
+                /* Delaying the animation like this is not ideal, but it is necessary
+                 * to allow some time for the window to paint itself for the first time.
+                 *
+                 * Mutter doesn't seem to have any kind of notification for when we first
```
cosimoc: Mutter has a `first-frame` signal that it emits to track this specific case. Not sure if it's helpful here.

js/ui/windowManager.js
```diff
@@ -571,6 +586,44 @@ const TilePreview = new Lang.Class({
...
 });
 
+const EnableBackfaceCullingEffect = new Lang.Class({
+    Name: 'EnableBackfaceCullingEffect',
+    Extends: Clutter.Effect,
+    _init : function(params) {
+        this.parent(params);
+    },
+    vfunc_pre_paint: function() {
+        this.get_actor().set_cull_back_face(true);
```
cosimoc: The `pre_paint` vfunc should return a boolean

js/ui/windowManager.js
```diff
@@ -571,6 +586,44 @@ const TilePreview = new Lang.Class({
...
+    },
+    vfunc_post_paint: function() {
+        this.get_actor().set_cull_back_face(true);
```
cosimoc: Are you sure you didn't want to set this back to `false` here? Regardless, a comment on why this effect is needed at all would be nice...

js/ui/windowManager.js
```diff
@@ -717,6 +784,126 @@ const WindowManager = new Lang.Class({
...
+
+                this._rotateInActors.push(animationSpec.dst.window);
+                this._rotateOutActors.push(animationSpec.src.window);
+
+                let waitingRotateTimeout = GLib.timeout_add_seconds(GLib.PRIORITY_DEFAULT, 1, Lang.bind(this, function() {
+                    /* Tween both windows in a rotation animation at the same time
+                     * with backface culling enabled on both. This will allow for
+                     * a smooth transition. */
+                    Tweener.addTween(animationSpec.src.window, {
+                        "rotation-angle-y": 180,
```
cosimoc: `rotation_angle_y` (multiple occurrences)

js/ui/windowManager.js
```diff
@@ -717,6 +784,126 @@ const WindowManager = new Lang.Class({
...
+                        time: WINDOW_ANIMATION_TIME * 4,
+                        transition: 'easeOutQuad',
+                        onComplete: function() {
+                            this._rotateOutCompleted(animationSpec.src.window);
+                        },
+                        onCompleteScope: this,
+                        onCompleteParams: []
```
cosimoc: You can remove this -- it's the default (multiple occurrences)

js/ui/windowManager.js
```diff
@@ -717,6 +784,126 @@ const WindowManager = new Lang.Class({
...
+                            this._rotateInCompleted(animationSpec.dst.window);
+                        },
+                        onCompleteScope: this,
+                        onCompleteParams: []
+                    });
+
+                    /* Gently fade the window in, this will paper over
+                     * any artifacts from shadows and the like */
+                    Tweener.addTween(animationSpec.dst.window, {
+                        "opacity": 255,
```
cosimoc: No quotes needed

js/ui/windowManager.js
```diff
@@ -717,6 +784,126 @@ const WindowManager = new Lang.Class({
...
+                animationSpec.dst.window.get_meta_window().move_resize_frame(false,
+                                                                             dst_geometry.x,
+                                                                             dst_geometry.y,
+                                                                             dst_geometry.width,
+                                                                             dst_geometry.height);
+
+                this._rotateInActors.push(animationSpec.dst.window);
+                this._rotateOutActors.push(animationSpec.src.window);
+
+                let waitingRotateTimeout = GLib.timeout_add_seconds(GLib.PRIORITY_DEFAULT, 1, Lang.bind(this, function() {
```
cosimoc: Is the timeout here due to the comment above? One second seems like way too long to wait for a first draw...

js/ui/windowManager.js
```diff
@@ -717,6 +784,126 @@ const WindowManager = new Lang.Class({
...
+                if (animationSpec[key].window == null) {
+                    if (animationSpec[key].pid == pid) {
+                        animationSpec[key].window = window;
+                    } else {
+                        unsatisfiedPids++;
+                    }
+                }
+            });
+            
+            if (unsatisfiedPids == 0) {
```
cosimoc: Prefer an early return when this is not zero

js/ui/windowManager.js
```diff
@@ -1096,6 +1283,11 @@ const WindowManager = new Lang.Class({
         }));
 
         let isSplashWindow = Shell.WindowTracker.is_speedwagon_window(window);
+        
+        if (this._updateReadyRotateAnimationsWith(actor)) {
+            shellwm.completed_map(actor);
+            return;
+        }
```
cosimoc: You should probably move this either after the `(!isSplashWindow)` check

js/ui/windowManager.js
```diff
@@ -1211,6 +1403,22 @@ const WindowManager = new Lang.Class({
...
+            Tweener.removeTweens(actor);
+            actor.opacity = 255;
+            actor.rotation_angle_x = 0;
+        }
+    },
+
+    _rotateOutCompleted: function(actor) {
+        if (this._removeEffect(this._rotateOutActors, actor)) {
+            Tweener.removeTweens(actor);
+            actor.hide();
```
cosimoc: What happens when one wants to go back to the original application?

smspillaz: This isn't implemented yet.

I'm not actually sure about the best way to do that. I suppose it could be that when the builder-window is closed, we rotate back to the original application. Either that, or we need some sort of mechanism to tell the shell to go back to the original application and hide builder.

I suspect that's going to be a fair bit of work, so I'd prefer to deal with that in a follow up.

js/ui/windowManager.js
smspillaz: I've noticed that builder seems to take quite some time before it first draws the window and one second seems to be the amount of time I need to wait before it is usually ready on my machine. On other machines it could be slower.

I'm not really sure about the best way of dealing with this. Digging around mutter, it seems like there doesn't seem to be any tracking of when a window was first damaged (which is technically when the pixmap contents are first defined).

js/ui/windowManager.js
smspillaz: Ah, yes, that would be very helpful.

js/ui/windowManager.js
smspillaz: See above, `first-frame` may be helpful here.

js/ui/windowManager.js
smspillaz: Hmm, point taken on the first bit that I need to clean up my use of terminology here. However, I don't understand the second part - this function just returns a struct with the first matching window actor, its size and pid, which we use later in detecting when a rotate animation can be performed.

js/ui/windowManager.js
smspillaz: This worked great! Thanks so much for the tip!

js/ui/windowManager.js
smspillaz: Ah, I think I understand the point of this comment now - the whole point was about the inconsistent use of terminology. I'm not sure where that came from (or why it even worked), but in any event I've just consistently referred to the property as `window` to avoid confusion.

js/ui/windowManager.js
```diff
@@ -2,6 +2,7 @@
 
 const Cairo = imports.cairo;
 const Clutter = imports.gi.Clutter;
+const Cogl = imports.gi.Cogl;
```
cosimoc: Remove this

js/ui/windowManager.js
```diff
@@ -609,6 +648,17 @@ const WindowManager = new Lang.Class({
             this._unminimizeWindowDone(shellwm, actor);
             this._mapWindowDone(shellwm, actor);
             this._destroyWindowDone(shellwm, actor);
+            this._rotateInCompleted(actor);
+            this._rotateOutCompleted(actor);
+
+            if (actor._firstFrameConnection) {
+                actor.disconnect(actor._firstFrameConnection);
+                this._firstFrameConnections = this._firstFrameConnections.filter(function(timeout) {
+                    return timeout !== actor._firstFrameConnection;
```
cosimoc: Nitpick: these are not timeouts anymore.

js/ui/windowManager.js
```diff
@@ -717,6 +767,135 @@ const WindowManager = new Lang.Class({
...
+                /* Gently fade the window in, this will paper over
+                 * any artifacts from shadows and the like */
+                Tweener.addTween(animationSpec.dst.window, {
+                    opacity: 255,
+                    time: WINDOW_ANIMATION_TIME,
+                    transition: 'linear'
+                });
+
+                this._firstFrameConnections = this._firstFrameConnections.filter(function(timeout) {
+                    return timeout != animationSpec.dst.window._firstFrameConnection;
```
cosimoc: See above comment

js/ui/windowManager.js
```diff
@@ -717,6 +767,135 @@ const WindowManager = new Lang.Class({
...
+                });
+                animationSpec.dst.window.disconnect(animationSpec.dst.window._firstFrameConnection);
+                animationSpec.dst.window._firstFrameConnection = null;
+
+                return false;
+            }));
+
+            /* Save the timeout's id on the destination window and in a list too so we can
```
cosimoc: Not a timeout

js/ui/windowManager.js
cosimoc: I would like to make that explicit. In other words, there should always be a way to rotate back to the original application, even without closing builder necessarily.
But yeah, let's save that for another follow up.

smspillaz:
Comments addressed

js/ui/windowManager.js
smspillaz: I agree.

At least for now we can probably just expose another d-bus method to pick a PID and "rotate it back" if it was already rotated out. The actual detail of how this is invoked can be determined at some other period. Thoughts?

PR #169, opened 2016-09-20
Update to latest eos-coverage.m4

ptomato:
https://phabricator.endlessm.com/T10602


smspillaz:
CR Pass, though it would be nice to have this in a central place. Lets update them all for now though.

PR #170, opened 2016-09-26
shell-js: Remove usage of deprecated API

ptomato:
This removes all usage of SpiderMonkey API that is deprecated in mozjs24
and will be removed in mozjs31.

https://bugzilla.gnome.org/show_bug.cgi?id=742249
https://phabricator.endlessm.com/T13537

### DIFF
--- a/src/shell-js.cpp
+++ b/src/shell-js.cpp
@@ -27,7 +27,7 @@ shell_js_add_extension_importer (const char  *target_object_script,
                                  const char  *directory,
                                  GError     **error)
 {
-  jsval target_object;
+  JS::Value target_object;
   GList *contexts;
   JSContext *context;
   char *search_path[2] = { 0, 0 };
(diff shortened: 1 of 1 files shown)


src/shell-js.cpp
```diff
@@ -61,14 +61,15 @@ shell_js_add_extension_importer (const char  *target_object_script,
       goto out;
     }
 
-  if (!JSVAL_IS_OBJECT (target_object))
+  if (!target_object.isObjectOrNull ())
```
smspillaz: Should there be a check here to see if this is an object and not null?

src/shell-js.cpp
```diff
@@ -61,14 +61,15 @@ shell_js_add_extension_importer (const char  *target_object_script,
...
     {
       g_error ("shell_js_add_extension_importer: invalid target object");
       goto out;
     }
 
   search_path[0] = (char*)directory;
-  gjs_define_importer (context, JSVAL_TO_OBJECT (target_object), target_property, (const char **)search_path, FALSE);
+  gjs_define_importer (context, target_object.toObjectOrNull (),
```
smspillaz: See above - should this throw an error if it converts to null?

src/shell-js.cpp
ptomato: Well, this is exactly the equivalent of the code that was there before; `JSVAL_IS_OBJECT` and `..._TO_OBJECT` confusingly counted `null` as an object. I can definitely see changing it to `isObject()` and `toObject()`, but I thought I'd err on the side of not changing the behaviour.

Now that I think about it, though, it doesn't make sense to define an importer on `null` since `null` can't have properties.

cosimoc:
This was merged upstream too -- accepting.

PR #173, opened 2016-10-07
Disable password reset except on Endless hardware

mcatanzaro:
The password reset feature is only desired on Endless hardware, not on
the free downloadable images nor on OEM images. Image type is determined
by checking the eos-image-version xattr of /sysroot. A gsetting is
provided to override the default behavior. (Note that the gsetting has
to be set for the gdm user, not your own user account.)

https://phabricator.endlessm.com/T13460

### DIFF
--- a/js/gdm/loginDialog.js
+++ b/js/gdm/loginDialog.js
@@ -753,6 +753,28 @@ const LoginDialog = new Lang.Class({
         if (!this._ensureCustomerSupportData())
             return;
 
+        let policy = global.settings.get_enum('password-reset-allowed');
+
+        // Explicitly disabled by user.
+        if (policy == 0)
+            return;
+
+        // Default value: enabled only on nonfree images. Image is nonfree if
+        // the eos-image-version xattr of /sysroot begins with "eosnonfree-".
+        if (policy == -1) {
+            try {
+                let file = Gio.file_new_for_path('/sysroot');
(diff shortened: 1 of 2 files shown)


js/gdm/loginDialog.js
```diff
@@ -753,6 +753,24 @@ const LoginDialog = new Lang.Class({
         if (!this._ensureCustomerSupportData())
             return;
 
+        let settings = Gio.Settings.new('org.gnome.shell');
```
cosimoc: You should be able to use `global.settings`, no?

js/gdm/loginDialog.js
```diff
@@ -753,6 +753,24 @@ const LoginDialog = new Lang.Class({
...
+
+        // Explicitly disabled by user.
+        if (policy == 0)
+            return;
+
+        // Default value: enabled only on nonfree images. Image is nonfree if
+        // the eos-image-version xattr of /sysroot begins with "eosnonfree-".
+        if (policy == -1) {
+            let file = Gio.file_new_for_path('/sysroot');
+            let file_info = file.query_info('xattr::eos-image-version', Gio.FileQueryInfoFlags.NONE, null);
```
cosimoc: Nitpick: please use camelCase for variable names

cosimoc: This should really be in a `try/catch` block

PR #176, opened 2016-10-12
taskbar-pins: add eos-installer icon (when image-booted)

wjt:
In general, this will actually have no effect: the taskbar ignores
desktop files that don't exist, and the eos-installer.desktop file
generally does not exist.

But when we are booted from an installable image, eos-installer.desktop
will be magicked into place. This way, if the user opts to "Try Endless"
during the FBE, they can still run eos-installer without rebooting.

https://phabricator.endlessm.com/T12548


wjt:
Rejected.

PR #183, opened 2016-10-17
extensionUtils: Remove ShellJS library

ptomato:
You can define a new importer object by importing a subdirectory in GJS.
This is undocumented, but it is likely to at least hold until the whole
thing moves to ES6 modules, after which we'll be able to do this purely
in JS with Reflect.Loader.

Since this was the only thing the ShellJS library did, we can remove it
altogether.

This allows us to discontinue use of the gjs-internals-1.0 embedder API.

https://bugzilla.gnome.org/show_bug.cgi?id=772386
https://phabricator.endlessm.com/T13622


ptomato:
Don't merge this yet; neither Florian nor I tested it on more than one extension running at the same time.

PR #188, opened 2016-10-19
T12729 squashed

cosimoc:
https://phabricator.endlessm.com/T12729

### DIFF
--- a/js/ui/sessionMode.js
+++ b/js/ui/sessionMode.js
@@ -97,6 +97,30 @@ const _modes = {
             right: ['dateMenu', 'a11y', 'keyboard', 'volume', 'bluetooth',
                     'network', 'battery', 'socialBar', 'hotCornerIndicator']
         }
+    },
+
+    'user-mission': {
+        hasOverview: true,
+        showCalendarEvents: true,
+        allowSettings: true,
+        allowExtensions: true,
+        allowScreencast: true,
+        hasRunDialog: true,
+        hasWorkspaces: true,
+        hasWindows: true,
+        hasNotifications: true,
+        isLocked: false,
+        isPrimary: true,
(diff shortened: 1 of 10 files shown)


js/ui/popupMenu.js
```diff
@@ -776,6 +776,34 @@ const PopupUserMenuItem = new Lang.Class({
...
+            cr.setLineWidth(stippleWidth);
+            cr.stroke();
+            cr.$dispose();
+        }));
+    }
+});
+
 const PopupMenuBase = new Lang.Class({
     Name: 'PopupMenuBase',
     Abstract: true,
```
smspillaz: We may not need this commit anymore actually, but it might be a good candidate for upstreaming?

js/ui/sessionMode.js
```diff
@@ -153,6 +177,11 @@ const SessionMode = new Lang.Class({
             let primary = modes[global.session_mode] &&
                           modes[global.session_mode].isPrimary;
             let mode = primary ? global.session_mode : 'user';
+
+            if (mode == 'user' &&
+                global.settings.get_boolean('enable-mission-game'))
+                mode = 'user-mission';
```
smspillaz: Ah interesting - that's a handy way to enable and disable this stuff without too much invasive change. Though that said, I think the effect of this change will be that the mission game mode will only be enabled across login/logout (and we likely won't be able to apply the setting immediately, for instance, if the user decides they want to turn it off for a while).

We might be okay with that for now though, so if that's good, then this is fine by me to merge.

smspillaz:
This is generally in an LGTM state. We might want to drop the first commit since it isn't relevant to anything anymore.

cosimoc:
Dropped the first commit, merging now.

PR #196, opened 2016-11-08
Properly detect changes in .desktop files

aperezdc:
This fixes the shell not picking updates to the contents of the underlying `.desktop` files for application launchers, or when a `.desktop` file is overriden by a user-installed one under `~/.local/share/applications`.

A direct backport from the upstream patch was not possible, so this is an adaptation of the patch for GNOME Shell which fixes [bug #773636](https://bugzilla.gnome.org/show_bug.cgi?id=773636).

https://phabricator.endlessm.com/T14057

### DIFF
--- a/src/shell-app-system.c
+++ b/src/shell-app-system.c
@@ -177,12 +177,13 @@ get_new_desktop_app_info_from_app (ShellApp *app)
 }
 
 static gboolean
-app_info_changed (ShellApp *app, GDesktopAppInfo *info)
+app_info_changed (ShellApp *app, GDesktopAppInfo *desk_new_info)
 {
   GIcon *app_icon;
   GIcon *new_icon;
-  GAppInfo *app_info = G_APP_INFO (shell_app_get_app_info (app));
-  GAppInfo *new_info = G_APP_INFO (info);
+  GDesktopAppInfo *desk_app_info = shell_app_get_app_info (app);
+  GAppInfo *app_info = G_APP_INFO (desk_app_info);
+  GAppInfo *new_info = G_APP_INFO (desk_new_info);
 
   if (!app_info)
     return TRUE;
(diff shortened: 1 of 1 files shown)


src/shell-app-system.c
```diff
@@ -177,20 +177,35 @@ get_new_desktop_app_info_from_app (ShellApp *app)
...
+           strcmp (g_app_info_get_executable (app_info),
+                   g_app_info_get_executable (new_info)) == 0 &&
+           g_strcmp0 (g_app_info_get_commandline (app_info),
+                      g_app_info_get_commandline (new_info)) == 0 &&
+           strcmp (g_app_info_get_name (app_info),
+                   g_app_info_get_name (new_info)) == 0 &&
+           strcmp (g_app_info_get_display_name (app_info),
+                   g_app_info_get_display_name (new_info)) == 0 &&
+           g_strcmp0 (g_app_info_get_description (app_info),
+                      g_app_info_get_description (new_info)) == 0);
```
smspillaz: Ouch, this is a huge return statement 😜

I wonder if it is worth refactoring this into separate functions and upstreaming the resulting patch (even if it is slightly different). Also, there's inconsistency here between using `strcmp` and `g_strcmp0` - probably one or the other should be used.

(Either that or it should probably be fixed in GIO with a `g_app_info_equal` function).

aperezdc:
PR updated with the changes suggested in the review for the [GNOME Shell version of the patch](https://bugzilla.gnome.org/show_bug.cgi?id=773636#c13). I have also submitted a [bug report + patch to GIO](https://bugzilla.gnome.org/show_bug.cgi?id=774302) to get this into `g_app_info_equal()`, but I think it is worth it to have this merged now to have the EOS Shell detecting application changes properly, and if GIO gets the fix, switch the comparison to use `g_app_info_equal()` later on, when we get an updated GLib. WDYT?

PR #198, opened 2016-11-11
Backport gtk notifications from upstream and add badges to the desktop icons

krnowak:
https://phabricator.endlessm.com/T13823

### DIFF
--- a/js/ui/appDisplay.js
+++ b/js/ui/appDisplay.js
@@ -31,6 +31,7 @@ const Tweener = imports.ui.tweener;
 const Workspace = imports.ui.workspace;
 const Params = imports.misc.params;
 const Util = imports.misc.util;
+const MessageTray = imports.ui.messageTray;
 
 const MAX_APPLICATION_WORK_MILLIS = 75;
 const MENU_POPUP_TIMEOUT = 600;
@@ -1820,6 +1821,21 @@ const AppFolderPopup = new Lang.Class({
 });
 Signals.addSignalMethods(AppFolderPopup.prototype);
 
+const AppIconSourceActor = new Lang.Class({
+    Name: 'AppIconSourceActor',
+    Extends: MessageTray.SourceActor,
+
+    _init: function(source, size) {
+        this.parent(source, size);
+        this.setIcon(new St.Bin());
+    },
+
+    _shouldShowCount: function() {
+        // Always show the counter when there's at least one notification
+        return this.source.count > 0;
+    }
+});
+
 const AppIcon = new Lang.Class({
     Name: 'AppIcon',
     Extends: ViewIcon,
@@ -1829,6 +1845,7 @@ const AppIcon = new Lang.Class({
         this._name = this.app.get_name();
 
         iconParams = Params.parse(iconParams, { createIcon: Lang.bind(this, this._createIcon),
+                                                createExtraIcons: Lang.bind(this, this._createExtraIcons),
                                                 editableLabel: true,
                                                 shadowAbove: true },
                                   true);
@@ -1839,6 +1856,9 @@ const AppIcon = new Lang.Class({
                                                 Lang.bind(this,
                                                           this._onStateChanged));
         this._onStateChanged();
+
+        Main.notificationDaemon.gtk.connect('new-gtk-notification-source',
+                                            Lang.bind(this, this._onNewGtkNotificationSource));
     },
 
     _onDestroy: function() {
@@ -1852,6 +1872,27 @@ const AppIcon = new Lang.Class({
 
     _createIcon: function(iconSize) {
         return this.app.create_icon_texture(iconSize);
+    },
+
+    _createExtraIcons: function(iconSize) {
+        if (!this._notificationSource)
+            return [];
+
+        let sourceActor = new AppIconSourceActor(this._notificationSource, iconSize);
+        return [sourceActor.actor];
+    },
+
+    _onNewGtkNotificationSource: function(daemon, source) {
+        if (source.app != this.app)
+            return;
+
+        this._notificationSource = source;
+        this._notificationSource.connect('destroy', Lang.bind(this, function() {
+            this._notificationSource = null;
+            this.icon.reloadIcon();
+        }));
+
+        this.icon.reloadIcon();
     },
 
     _onStateChanged: function() {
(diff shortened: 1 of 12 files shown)


js/ui/appDisplay.js
```diff
@@ -1820,12 +1821,96 @@ const AppFolderPopup = new Lang.Class({
...
+Signals.addSignalMethods(AppIconSource.prototype);
+
 const AppIcon = new Lang.Class({
     Name: 'AppIcon',
     Extends: ViewIcon,
 
     _init : function(app, iconParams, params) {
         this.app = app;
+        this._iconSource = new AppIconSource(this);
+        this._sourceActor = null;
```
krnowak: Hmm, dunno. We don't return `this._sourceActor` from the `_createIcon` function, but rather only `this._sourceActor.actor`. What would happen to `this._sourceActor` if it is not a private member and then gets out of scope? Would it be GC'd or would it survive because the source actor binds itself to some signals in its `_init` function? If the former, then we need to keep it alive by having it as a private member, I think.

js/ui/notificationDaemon.js
```diff
@@ -876,7 +876,8 @@ const GtkNotificationDaemonAppSource = new Lang.Class({
             delete this._notificationsDestroyIds[notificationId];
         }));
         this._notifications[notificationId] = notification;
-
+        // TODO(krnowak): set notification.isMusic
+        // TODO(krnowak): What's up with the trayicon?
```
krnowak: https://phabricator.endlessm.com/T14198

js/ui/notificationDaemon.js
```diff
@@ -10,6 +10,7 @@ const Lang = imports.lang;
 const Shell = imports.gi.Shell;
 const Mainloop = imports.mainloop;
 const St = imports.gi.St;
+const Signals = imports.signals;
```
krnowak: Sorted, emacs moved some other lines too.

js/ui/appDisplay.js
```diff
@@ -1865,8 +1865,10 @@ const AppIconSource = new Lang.Class({
         let thisAppId = this._desktopLessAppId(this._appIcon.app.get_id());
         let sourceAppId = this._desktopLessAppId(appId);
 
+        log('the app icon source for ' + thisAppId + ' got new gtk notification source for app ' + sourceAppId);
```
krnowak: Removed.

js/ui/iconGrid.js
```diff
@@ -317,7 +317,7 @@ const BaseIcon = new Lang.Class({
             this.createIcon = params.createIcon;
         this._setSizeManually = params.setSizeManually;
 
-        this.icon = null;
+        this.icons = [];
```
krnowak: Yes, this is a better idea. Added the `createExtraIcons()` method just for that.

krnowak:
Updated.

js/ui/notificationDaemon.js
```diff
@@ -825,3 +717,315 @@ const Source = new Lang.Class({
...
+
+const GtkNotificationDaemonNotification = new Lang.Class({
+    Name: 'GtkNotificationDaemonNotification',
+    Extends: MessageTray.Notification,
+
+    _init: function(source, notification) {
+        this.parent(source);
+        this._serialized = GLib.Variant.new('a{sv}', notification);
+
+        let { "title": title,
```
erikos: The other places in the shell are using without "". E.g. js/ui/workspace.js:770:        let { rows: rows, scale: scale } = layout;

krnowak: This is the backported patch, I don't think we will want to change that.

krnowak:
Updated again, forgot to check if it actually runs after my changes. :/

So, I need to test it a bit more (that is - I need to build again the newest bloatpad with the gtk gnotification thing to test the badges) and to check whether the exception thrown when using fdo notifications is my fault. The exception happens when the notification is shown, then I put the mouse pointer over it, so the notification expands. Then, when I move the mouse pointer away from the notification, the notification abruptly disappears and the eos-shell logs some exception.

js/ui/appDisplay.js
cosimoc: I am not completely sure, I guess we should test it :-)
But in general, I have not really encountered cases where GC gets in the way, so I'd be surprised if it was an issue in practice.

cosimoc:
@krnowak I can't reproduce that exception you mentioned using the regular branch of eos-desktop, so I can only assume it's introduced by this branch.
Please try to fix it before we can merge this.

js/ui/appDisplay.js
cosimoc: However, don't spend too much time on that... I am actually OK with it using a private member like you have it.

js/ui/notificationDaemon.js
cosimoc: Yeah, we don't touch backported code.

krnowak:
I tested this thing on the latest eos-shell package and I'm getting this:
```
(eos-shell:1124): Gjs-WARNING **: JS ERROR: Error calling onComplete: TypeError: this._notification is null
MessageTray<._hideNotificationCompleted@resource:///org/gnome/shell/ui/messageTray.js:2596
wrapper@resource:///org/gnome/gjs/modules/lang.js:178
MessageTray<._tweenComplete@resource:///org/gnome/shell/ui/messageTray.js:2346
wrapper@resource:///org/gnome/gjs/modules/lang.js:178
_addHandler/params[name]@resource:///org/gnome/shell/ui/tweener.js:89
_callOnFunction@resource:///org/gnome/gjs/modules/tweener/tweener.js:203
_updateTweenByIndex@resource:///org/gnome/gjs/modules/tweener/tweener.js:333
_updateTweens@resource:///org/gnome/gjs/modules/tweener/tweener.js:345
_onEnterFrame@resource:///org/gnome/gjs/modules/tweener/tweener.js:360
_emit@resource:///org/gnome/gjs/modules/signals.js:124
ClutterFrameTicker<._onNewFrame@resource:///org/gnome/shell/ui/tweener.js:206
wrapper@resource:///org/gnome/gjs/modules/lang.js:178
ClutterFrameTicker<._init/<@resource:///org/gnome/shell/ui/tweener.js:181
```

So, not my fault and probably harmless (the notification disappears abruptly instead of sliding down). There is still this freezing button thing left.

PR #201, opened 2016-11-22
A few Chinese edits

Qiangong2:
Traditional Chinese additions


cosimoc:
@Qiangong2 Thanks for the updated translations!
@rshuler I believe we should update the strings directly in Transifex as opposed to here, right? Could you do that?

rshuler:
This is great -- thanks!  Actually, what I can do is merge the pull request and then push the translations into Phabricator.  I'll do that now, and these translations will show up starting with our 3.1.0 release.

PR #208, opened 2016-12-08
Clean up ShellGlobal before exit

ptomato:
https://phabricator.endlessm.com/T14460

### DIFF
--- a/src/gnome-shell-extension-prefs.c
+++ b/src/gnome-shell-extension-prefs.c
@@ -28,6 +28,7 @@ main (int argc, char **argv)
     {
       g_message("Failed to defined ARGV: %s", error->message);
       g_error_free (error);
+      g_object_unref (context);
 
       return 1;
     }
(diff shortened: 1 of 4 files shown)


src/gnome-shell-extension-prefs.c
```diff
@@ -28,6 +28,7 @@ main (int argc, char **argv)
     {
       g_message("Failed to defined ARGV: %s", error->message);
       g_error_free (error);
+      g_object_unref (context);
```
smspillaz: I'm not sure if we're using it here, but if we have `g_autoptr` it seems like this is a good candidate for its use.

ptomato:
@smspillaz Cosimo merged it before I got a chance to address your comments ... in fact I did try deleting `window.global` in one iteration of this code, but you can't wait around for a GC to finish so there's no guarantee that the object is actually gone by the time you want to destroy the object.

PR #217, opened 2017-01-10
CodeView: listen to overview hiding and showing to hide overlay buttons

erikos:
When the desktop is shown and the windows are hidden and vice
versa we need to make sure the overlay buttons are hidden and
shown again in those cases. The overview emits signals for
those cases used at other places in the shell already.

In the handlers we do not try to figure out the current active
session window and hide/show the existing buttons in any case.

https://phabricator.endlessm.com/T14811

### DIFF
--- a/js/ui/codingManager.js
+++ b/js/ui/codingManager.js
@@ -134,6 +134,8 @@ const CodingManager = new Lang.Class({
         session.sizeChangedIdApp = window.connect('size-changed', Lang.bind(this, this._updateAppSizeAndPosition, session));
 
         session.windowsRestackedId = Main.overview.connect('windows-restacked', Lang.bind(this, this._windowRestacked, session));
+        session.overviewHidingId = Main.overview.connect('hiding', Lang.bind(this, this._overviewHiding, session));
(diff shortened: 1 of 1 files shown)


js/ui/codingManager.js
```diff
@@ -462,6 +472,18 @@ const CodingManager = new Lang.Class({
...
+        session.buttonApp.show();
+        if (session.buttonBuilder)
+            session.buttonBuilder.show();
+    },
+
+    _overviewShowing: function(overview, session) {
+        session.buttonApp.hide();
+        if (session.buttonBuilder)
+            session.buttonBuilder.hide();
+    },
```
smspillaz: I know that there's a few other places where we have to do this as well - is it possible to combine this all into one function? (e.g. `showCodeviewButtons` / `hideCodeviewButtons`) and connect the signals to those or thunks which call those? There's a bit of logic here in terms of checking that `buttonBuilder` is truthy that is getting repeated.

erikos: Thanks for the review Sam. I actually looked at the code and while the feeling is that there is some replication, there is not that much in the end. I would keep it that way without refactoring.

smspillaz:
Okay. Lets merge as is then.

smspillaz:
LGTM.

PR #219, opened 2017-01-19
userWidget: Fix leak of UserWidget causing crashes

pwithnall:
In the two places the UserWidget was used (the login dialog and unlock
dialog), a UserWidget was constructed, then its `actor` property was
added to the UI and used. The actor was destroyed once done with — but
the UserWidget itself was never destroyed (by calling its `destroy()`
method).

This resulted in a couple of signals remaining connected to the D-Bus
proxy for the org.freedesktop.login1.User instance which the widget was
representing — notably its `changed` signal. This is emitted (amongst
other times) when the list of sessions the user is in changes, which can
happen if the user logs in via SSH while also logged in graphically, for
example.

Consequently, if the lock screen had been used (and hence a UserWidget
created for the UnlockDialog), then the user’s session list changed, the
callback for `changed` was called with a set of widgets which had been
destroyed. It would try and set the user’s username on them, and crash
due to using the dangling widget pointers.

Fix that by explicitly destroying the UserWidget instances at
appropriate times in the lifecycles of LoginDialog and UnlockDialog.

Signed-off-by: Philip Withnall <withnall@endlessm.com>

https://phabricator.endlessm.com/T12270

---

I’ve tested this by monkey-patching it in a gnome-shell extension, rather than by rebuilding the shell with it; but the end result should be the same.

### DIFF
--- a/js/gdm/loginDialog.js
+++ b/js/gdm/loginDialog.js
@@ -1078,6 +1078,8 @@ const LoginDialog = new Lang.Class({
         this._promptLoginHint.hide();
 
         this._promptUser.set_child(null);
+        this._userWidget.destroy();
+        this._userWidget = null;
 
         this._updateSensitivity(true);
         this._promptEntry.set_text('');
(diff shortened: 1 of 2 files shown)


pwithnall:
Hold off on reviewing this; it looks like this was fixed upstream 4 years ago and I can backport the fix.

pwithnall:
Yes, the [upstream patch](https://bugzilla.gnome.org/show_bug.cgi?id=700807) works, and is a better approach. Abandoning this PR in favour of PR #220.

PR #221, opened 2017-01-19
T15172 fix json parsing

pwithnall:
A couple of fixes for JSON parsing in `internetSearch.js`.

https://phabricator.endlessm.com/T15172

### DIFF
--- a/js/ui/internetSearch.js
+++ b/js/ui/internetSearch.js
@@ -64,17 +64,33 @@ const InternetSearchProvider = new Lang.Class({
         }
 
         let root = parser.get_root().get_object();
-        let searchProviderData = root.get_object_member('default_search_provider_data');
+
+        let searchProviderDataNode = root.get_member('default_search_provider_data');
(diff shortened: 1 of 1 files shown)


js/ui/internetSearch.js
```diff
@@ -55,23 +55,42 @@ const InternetSearchProvider = new Lang.Class({
...
             return null;
         }
 
-        return templateUrlData.get_string_member('short_name');
+        let shortNameNode = templateUrlData.get_member('short_name');
+        if (!shortNameNode || shortNameNode.get_node_type() != Json.NodeType.VALUE) {
+            return null;
+        }
+
+        return shortNameNode.get_string();
```
smspillaz: Is there any reason why we're using JSON-GLib here as opposed to just using JavaScript's native JSON.parse method? It seems like this results in overly verbose checks.

cosimoc:
Thanks @pwithnall I saw this and meant to fix it at some point :)
I can't remember either off-hand why I ended up using json-glib, but your PR makes sense to me without changing the existing code too much.

PR #224, opened 2017-01-24
CodeView: Desktop hide/show, handle case where Builder and app are vi…

erikos:
…sible

The coding session consists of two windows, the GNOME Builder
one and the application one, each window has an overlay button
respectively. When we hide the Desktop we only want to show
the one that is associated with the currently focused window.

Follow-up of: cd6fa28429934e2b877967af6085a1d68326c148

https://phabricator.endlessm.com/T14811

### DIFF
--- a/js/ui/codingManager.js
+++ b/js/ui/codingManager.js
@@ -473,9 +473,15 @@ const CodingManager = new Lang.Class({
     },
 
     _overviewHiding: function(overview, session) {
-        session.buttonApp.show();
-        if (session.buttonBuilder)
+        let focusedWindow = global.display.get_focus_window();
+        if (!focusedWindow)
+            return;
+
+        // we only need to verify a Builder window exist
+        if (session.actorBuilder && session.actorBuilder.meta_window === focusedWindow)
             session.buttonBuilder.show();
+        else if (session.actorApp.meta_window === focusedWindow)
+            session.buttonApp.show();
     },
 
     _overviewShowing: function(overview, session) {


js/ui/codingManager.js
```diff
@@ -473,9 +473,19 @@ const CodingManager = new Lang.Class({
...
+        let focusedWindow = global.display.get_focus_window();
+        if (!focusedWindow)
+            return;
+
+        let builderWindow = null;
+        if (session.actorBuilder)
+            builderWindow = session.actorBuilder.meta_window;
+
+        if (builderWindow && builderWindow === focusedWindow)
+             session.buttonBuilder.show();
```
smspillaz: nit: indentation

js/ui/codingManager.js
```diff
@@ -473,9 +473,19 @@ const CodingManager = new Lang.Class({
...
-            session.buttonBuilder.show();
+        let focusedWindow = global.display.get_focus_window();
+        if (!focusedWindow)
+            return;
+
+        let builderWindow = null;
+        if (session.actorBuilder)
+            builderWindow = session.actorBuilder.meta_window;
+
+        if (builderWindow && builderWindow === focusedWindow)
```
smspillaz: Can just compare against `session.actorBuilder.window` since `focusedWindow` is guaranteed to be non-null

smspillaz:
LGTM thanks

js/ui/codingManager.js
```diff
@@ -473,9 +473,14 @@ const CodingManager = new Lang.Class({
...
     _overviewHiding: function(overview, session) {
-        session.buttonApp.show();
-        if (session.buttonBuilder)
+        let focusedWindow = global.display.get_focus_window();
+        if (!focusedWindow)
+            return;
+
+        if (session.actorBuilder && session.actorBuilder.meta_window === focusedWindow)
             session.buttonBuilder.show();
+        else if (session.actorApp.meta_window === focusedWindow)
```
smspillaz: is `session.actorApp` nullable like `session.actorBuilder` above? If not, please add a comment.

smspillaz:
Spoke too soon - one more comment (but very minor)

PR #226, opened 2017-01-26
Hide speedwagon when the focused app changes

joaquimrocha:
If there is an issue with launching an application for which a
speedwagon has been shown, and that application only reports that it
is starting, then the speedwagon stays on forever, showing on top of
other apps which hinders UX.

This is the case for Google Chrome when it is not installed. In that
case, Google Chrome's icon launches the App Center instead, but since
the browser's own window is never shown -- so there's no state
transition for the shell to hide the speedwagon -- and the speedwagon
stays on forever.

This keeps the previous behavior of managing the speedwagon based on
its app's state but also hides it if the focused app is changed and
does not match the speedwagon's one.

https://phabricator.endlessm.com/T15129

### DIFF
--- a/js/ui/appActivation.js
+++ b/js/ui/appActivation.js
@@ -43,6 +43,12 @@ const AppActivationContext = new Lang.Class({
         this._appActivationTime = 0;
 
         this._hiddenWindows = [];
+
+        this._appSystem = Shell.AppSystem.get_default();
+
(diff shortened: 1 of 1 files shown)


cosimoc:
@joaquimrocha Would this not always clear the speedwagon as soon as the user presses alt-tab? I don't think that's what we want.

PR #228, opened 2017-01-28
appFavorites: handle automatic google-chrome integration

rshuler:
Similar to what we did for the icon grid, we want the default favorites
(e.g., the taskbar pins) to substitute chrome for chromium if we have
the integration enabled, and we save the modified settings so that
the configuration remains after chrome has been installed (which clears
the gsetting for the chrome initial setup integration).

https://phabricator.endlessm.com/T15296


rshuler:
Merging my own pull request to get the 3.1.1 release builds started tonight, but I'll leave the task in CR ready.

PR #229, opened 2017-02-08
iconGridLayout: add image default override option

rshuler:
By adding a JSON file in /var/eos-image-defaults/icon-grid,
the default icon grid for a locale can be overridden.
For instance, this allows the image builder to specify
having Chrome on the desktop by default rather than Chromium.

Follows the same syntax and locale-matching of the defaults in
/usr/share/eos-shell-content/icon-grid-defaults.

If there is any image default override that matches the user's locale,
it will take priority over any base default, even if the base default
is a more specific match to the user's locale.

Any image default prepending or appending is still applied
after this override.

https://phabricator.endlessm.com/T15324

### DIFF
--- a/js/ui/iconGridLayout.js
+++ b/js/ui/iconGridLayout.js
@@ -22,6 +22,9 @@ const FOLDER_DIR_NAME = 'desktop-directories';
 const DEFAULT_CONFIGS_DIR = Config.DATADIR + '/eos-shell-content/icon-grid-defaults';
 const DEFAULT_CONFIG_NAME_BASE = 'icon-grid';
 
+const OVERRIDE_CONFIGS_DIR = Config.LOCALSTATEDIR + '/eos-image-defaults/icon-grid';
+const OVERRIDE_CONFIG_NAME_BASE = 'icon-grid';
+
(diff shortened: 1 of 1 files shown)


js/ui/iconGridLayout.js
```diff
@@ -148,6 +158,7 @@ const IconGridLayout = new Lang.Class({
     _getDefaultIcons: function() {
         let mergedJson = this._mergeJsonStrings(
             this._loadConfigJsonString(DEFAULT_CONFIGS_DIR, DEFAULT_CONFIG_NAME_BASE),
+            this._loadConfigJsonString(OVERRIDE_CONFIGS_DIR, OVERRIDE_CONFIG_NAME_BASE),
```
cosimoc: I am not sure this will always do what you want; `mergeJsonStrings()` will create an "union" of all the settings specified in the various files, so if e.g. the `/usr` config for a language specifies a folder that is not in the `/var` config, the end result is that such folder will be in the merged JSON.
Instead, the code should load either the JSON from `/usr` or the one in `/var`, excluding the one in `/usr` in the latter case, then apply the append/prepend overrides to the result.

cosimoc: Talked with @rshuler and did not understand how this worked. Merged now.

PR #231, opened 2017-02-14
CodeView: use D-Bus activation for starting up the gnome-builder

krnowak:
https://phabricator.endlessm.com/T14628

### DIFF
--- a/js/ui/codingManager.js
+++ b/js/ui/codingManager.js
@@ -13,7 +13,6 @@ const St = imports.gi.St;
 
 const Main = imports.ui.main;
 const Tweener = imports.ui.tweener;
-const Util = imports.misc.util;
 
 const WINDOW_ANIMATION_TIME = 0.25;
 const WATCHDOG_TIME = 30000; // ms
@@ -247,8 +246,28 @@ const CodingManager = new Lang.Class({
         }
     },
 
+    _startBuilderForFlatpak: function(loadFlatpakValue) {
+        let params = new GLib.Variant('(sava{sv})', ['load-flatpak', [new GLib.Variant('s', loadFlatpakValue)], {}]);
+        Gio.DBus.session.call('org.gnome.Builder',
+                              '/org/gnome/Builder',
+                              'org.gtk.Actions',
+                              'Activate',
+                              params,
+                              null,
+                              Gio.DBusCallFlags.NONE,
+                              GLib.MAXINT32,
+                              null,
+                              function (conn, result) {
+                                  try {
+                                      conn.call_finish(result);
+                                  } catch (e) {
+                                      logError(e, 'Failed to start gnome-builder');
+                                  }
+                              });
+    },
+
     _switchToBuilder: function(actor, event, session) {
-        function constructCommand(appManifest) {
+        function constructLoadFlatpakValue(appManifest) {
             // add an app_id_override to the manifest to load
             return appManifest.get_path() + '+' + session.actorApp.meta_window.get_flatpak_id() + '.Coding';
         }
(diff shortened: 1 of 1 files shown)


js/ui/codingManager.js
```diff
@@ -267,7 +267,23 @@ const CodingManager = new Lang.Class({
             this._watchdogId = Mainloop.timeout_add(WATCHDOG_TIME,
                                                     Lang.bind(this, this._watchdogTimeout));
 
-            Util.trySpawn(['flatpak', 'run', 'org.gnome.Builder', '-s', '-m', constructCommand(appManifest)]);
```
erikos: const Util = imports.misc.util;

We can remove that import then.

erikos:
Ok, so just had a quick test as well, works fine. So, a separate function would be nice as Sam suggested and the removal of the unused util import.

js/ui/codingManager.js
```diff
@@ -267,7 +267,23 @@ const CodingManager = new Lang.Class({
...
+            let params = new GLib.Variant('(sava{sv})', ['load-flatpak', [new GLib.Variant('s', constructCommand(appManifest))], {}]);
+            Gio.DBus.session.call('org.gnome.Builder',
+                                  '/org/gnome/Builder',
+                                  'org.gtk.Actions',
+                                  'Activate',
+                                  params,
+                                  null,
+                                  Gio.DBusCallFlags.NONE,
+                                  -1,
```
cosimoc: You should pass `G_MAXINT` here to avoid timeouts

js/ui/codingManager.js
```diff
@@ -267,7 +267,23 @@ const CodingManager = new Lang.Class({
...
+                                  null,
+                                  function (conn, result) {
+                                      try {
+                                          conn.call_finish(result);
+                                      } catch (e) {
+                                          log('Failed to start gnome-builder: ' + e.toString());
```
cosimoc: Use `logError(e, 'message')`

js/ui/codingManager.js
krnowak: Finding javascript's counterpart of the C `G_MAXINT` macro was a bit challenging…

js/ui/codingManager.js
krnowak: Done.

krnowak:
Refactored to a separate function, but I haven't tested it - somehow the button for rotating the app stopped showing up in the gnome weather application…

erikos:
Thanks Krzesimir. Looking good and tested here as working. I would keep the helper function to construct the command though.
```

diff --git a/js/ui/codingManager.js b/js/ui/codingManager.js
index d923eef..a521719 100644
--- a/js/ui/codingManager.js
+++ b/js/ui/codingManager.js
@@ -246,10 +246,13 @@ const CodingManager = new Lang.Class({
         }
     },
 
-    _startBuilder: function(manifestPath, flatpakId) {
-        // add an app_id_override to the manifest to load
-        let flagValue = manifestPath + '+' + flatpakId + '.Coding';
-        let params = new GLib.Variant('(sava{sv})', ['load-flatpak', [new GLib.Variant('s', flagValue)], {}]);
+    _startBuilder: function(appManifest, flatpakId) {
+        function constructCommand(appManifest) {
+            // add an app_id_override to the manifest to load
+            return appManifest.get_path() + '+' + flatpakId + '.Coding';
+        }
+
+        let params = new GLib.Variant('(sava{sv})', ['load-flatpak', [new GLib.Variant('s', constructCommand(appManifest))], {}]);
         Gio.DBus.session.call('org.gnome.Builder',
                               '/org/gnome/Builder',
                               'org.gtk.Actions',
@@ -283,7 +286,7 @@ const CodingManager = new Lang.Class({
             this._watchdogId = Mainloop.timeout_add(WATCHDOG_TIME,
                                                     Lang.bind(this, this._watchdogTimeout));
 
-            this._startBuilder(appManifest.get_path(),
+            this._startBuilder(appManifest,
                                session.actorApp.meta_window.get_flatpak_id());
             animateBounce(session.buttonApp);
         } else {

```

PR #232, opened 2017-02-16
iconGridLayout: Provide default for missing icon grid files

ptomato:
This makes the default simply an empty icon grid, avoiding a crash if the
icon grid files are not present.

https://phabricator.endlessm.com/T15606

### DIFF
--- a/js/ui/iconGridLayout.js
+++ b/js/ui/iconGridLayout.js
@@ -128,14 +128,14 @@ const IconGridLayout = new Lang.Class({
     },
 
     _mergeJsonStrings: function(base, override, prepend, append) {
-        let baseNode = null;
+        let baseNode = {};
         let prependNode = null;
         let appendNode = null;
         // If any image default override matches the user's locale,
         // give that priority over the default from the base OS
         if (override) {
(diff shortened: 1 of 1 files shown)


js/ui/iconGridLayout.js
```diff
@@ -104,7 +104,7 @@ const IconGridLayout = new Lang.Class({
     },
 
     _loadConfigJsonString: function(dir, base) {
-        let jsonString = null;
+        let jsonString = '{}';
```
rshuler: Actually, this will break things, because the code in _mergeJsonStrings is relying on this to be null for the missing data (as there are two possible sources of this information -- one part of the core OS, and one provided by the image builder).  Instead, I think the correct change is in _mergeJsonStrings, something along the lines of the following...
From:
```
        if (override) {
            baseNode = JSON.parse(override);
        } else {
            baseNode = JSON.parse(base);
        }
```
To:
```
        if (override) {
            baseNode = JSON.parse(override);
        } else if (base) {
            baseNode = JSON.parse(base);
        } else {
            baseNode = JSON.parse('{}');
        }
```

PR #240, opened 2017-03-01
messageTray: Reset clickedSummaryItem on ungrab

cosimoc:
If we don't, we will pop up the summary again the next time
_updateState() is called.

https://bugzilla.gnome.org/show_bug.cgi?id=707600

https://phabricator.endlessm.com/T15855


krnowak:
Looks good and seems to work.

PR #243, opened 2017-03-02
T15536

GeorgesStavracas:
Implement the new desktop search entry style.

https://phabricator.endlessm.com/T15536

### DIFF
--- a/js/misc/util.js
+++ b/js/misc/util.js
@@ -7,6 +7,7 @@ const Shell = imports.gi.Shell;
 const St = imports.gi.St;
 
 const Config = imports.misc.config;
+const Json = imports.gi.Json;
 const Tweener = imports.ui.tweener;
 
 const FALLBACK_BROWSER_ID = 'chromium-browser.desktop';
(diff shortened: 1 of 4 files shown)


js/misc/util.js
```diff
@@ -268,3 +269,67 @@ function blockClickEventsOnActor(actor) {
...
+
+    let root = parser.get_root().get_object();
+
+    let searchProviderDataNode = root.get_member('default_search_provider_data');
+    if (!searchProviderDataNode || searchProviderDataNode.get_node_type() != Json.NodeType.OBJECT) {
+        /*
+         * Translators: this is the name of the search engine that shows in the
+         * Shell's desktop search entry.
+         */
+        return _("Google");
```
mcatanzaro: Can't you make this a constant, so you don't have to hardcode Google all over the place?

mcatanzaro:
Sorry I was a bit distracted during standup, but did you file a new bug for the issue that it's pulling the default search engine name from Chromium and not the default web browser? Doesn't this seem like a good time to handle that as well? I would check the default browser with xdg-mime, if it's Chromium use this code, if it's Google use this code but with a different patch to the settings, and otherwise just use the "search the internet" string.

We should also test to make sure that launching search results works when Firefox is the default browser. I suspect it might not right now. That can be a separate issue, though.

PR #244, opened 2017-03-02
T15681 Additional refactoring with button animations

erikos:
https://phabricator.endlessm.com/T15681

### DIFF
--- a/js/ui/codingManager.js
+++ b/js/ui/codingManager.js
@@ -706,10 +706,11 @@ const CodingSession = new Lang.Class({
         if (!focusedWindow)
             return;
 
+        if (!this.builder)
+            return;
+
         let appWindow = this.app.meta_window;
-        let builderWindow = null;
-        if (this.builder)
-            builderWindow = this.builder.meta_window;
+        let builderWindow = this.builder.meta_window;
 
         if (appWindow === focusedWindow) {
             if (builderWindow && this._state === STATE_BUILDER) {
@@ -728,9 +729,6 @@ const CodingSession = new Lang.Class({
             }
             return;
         }
-
-        if (!this.builder)
-            return;
 
         if (builderWindow === focusedWindow) {
             if (this._state === STATE_APP) {
--- a/js/ui/codingManager.js
+++ b/js/ui/codingManager.js
@@ -34,11 +34,11 @@ _CODING_APPS = [
     'org.gnome.Weather'
 ];
 
-function _isCodingApp (flatpakID) {
+function _isCodingApp(flatpakID) {
     return _CODING_APPS.indexOf(flatpakID) != -1;
 }
 
-function _isBuilder (flatpakID) {
+function _isBuilder(flatpakID) {
     return flatpakID === 'org.gnome.Builder';
 }
 
--- a/js/ui/codingManager.js
+++ b/js/ui/codingManager.js
@@ -618,11 +618,12 @@ const CodingSession = new Lang.Class({
         return [src, dst];
     },
 
-    _srcAndDstPairFromWindowActor: function(actor) {
-        let src = actor;
-        let dst = (actor === this.app ? this.builder : this.app);
-
-        return [src, dst];
+    _isActorFromSession: function(actor) {
+        if (actor === this.app && this.builder)
+            return this.builder;
+        else if (actor === this.builder && this.app)
+            return this.app;
+        return null;
     },
 
     _synchronizeWindows: function(window) {
@@ -639,17 +640,19 @@ const CodingSession = new Lang.Class({
         if (!this.builder)
             return;
 
-        let [src, dst] = this._srcAndDstPairFromWindowActor(actor);
-        dst.meta_window.minimize();
+        let toMini = this._isActorFromSession(actor);
+        if (toMini)
+            toMini.meta_window.minimize();
     },
 
-    _applyWindowUnminimizationState: function(shellwm, window) {
+    _applyWindowUnminimizationState: function(shellwm, actor) {
         // No synchronization if builder has not been set here
         if (!this.builder)
             return;
 
-        let [src, dst] = this._srcAndDstPairFromWindowActor(window);
-        dst.meta_window.unminimize();
+        let toUnMini = this._isActorFromSession(actor);
+        if (toUnMini)
+            toUnMini.meta_window.unminimize();
     },
 
     _windowsRestacked: function() {
--- a/js/ui/codingManager.js
+++ b/js/ui/codingManager.js
@@ -103,6 +103,26 @@ function _synchronizeMetaWindowActorGeometries(src, dst) {
                                       srcGeometry.height);
 }
 
+function _synchronizeViewSourceButtonToRectCorner(button, rect) {
+    button.set_position(rect.x + rect.width - BUTTON_OFFSET_X,
+                        rect.y + rect.height - BUTTON_OFFSET_Y);
+}
+
+function _createViewSourceButtonInRectCorner(rect) {
+    let button = new St.Bin({
+        style_class: 'view-source',
+        reactive: true,
+        can_focus: true,
+        x_fill: true,
+        y_fill: false,
+        track_hover: true
+    });
+    button.set_child(_createIcon());
+    _synchronizeViewSourceButtonToRectCorner(button, rect);
+    Main.layoutManager.addChrome(button);
+    return button;
+}
+
 const WindowTrackingButton = new Lang.Class({
     Name: 'WindowTrackingButton',
     Extends: GObject.Object,
@@ -126,16 +146,10 @@ const WindowTrackingButton = new Lang.Class({
     _init: function(params) {
         this.parent(params);
 
-        // Add button asset and set the child of this bin
-        this._button = new St.Bin({
-            style_class: 'view-source',
-            reactive: true,
-            can_focus: true,
-            x_fill: true,
-            y_fill: false,
-            track_hover: true
-        });
-        this._button.set_child(_createIcon());
+        // The button will be auto-added to the manager. Note that in order to
+        // remove this button, you will need to call eject() from outside
+        // this class or use removeChrome from within it.
+        this._button = _createViewSourceButtonInRectCorner(this.window.get_frame_rect());
         this._button.connect('button-press-event', Lang.bind(this, function() {
             this.emit('clicked');
         }));
@@ -167,13 +181,6 @@ const WindowTrackingButton = new Lang.Class({
         this._windowUnminimizedId = global.window_manager.connect(
             'unminimize', Lang.bind(this, this._show)
         );
-
-        // Do the first position update here
-        this._updatePosition();
-
-        // Add ourselves to the layout manager. Note that in order to
-        // remove this button, you will need to call eject
-        Main.layoutManager.addChrome(this._button);
     },
 
     eject: function() {
@@ -283,8 +290,7 @@ const WindowTrackingButton = new Lang.Class({
 
     _updatePosition: function() {
         let rect = this.window.get_frame_rect();
-        this._button.set_position(rect.x + rect.width - BUTTON_OFFSET_X,
-                                  rect.y + rect.height - BUTTON_OFFSET_Y);
+        _synchronizeViewSourceButtonToRectCorner(this._button, rect);
     },
 
     _showIfWindowVisible: function() {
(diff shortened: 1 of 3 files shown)


js/ui/codingManager.js
```diff
@@ -210,20 +206,72 @@ const WindowTrackingButton = new Lang.Class({
...
+        Tweener.addTween(this, {
+            rotation_angle_y: direction == Gtk.DirectionType.RIGHT ? 180 : -180,
+            time: WINDOW_ANIMATION_TIME * 4,
+            transition: 'easeOutQuad'
+        });
         Tweener.addTween(this, {
             opacity: 0,
-            time: WINDOW_ANIMATION_TIME / 2,
-            transition: 'linear',
+            time: WINDOW_ANIMATION_TIME * 2,
```
smspillaz: Any reason to change this from  / and * and from `linear` to `easeOutQuad`?

js/ui/codingManager.js
```diff
@@ -210,20 +206,72 @@ const WindowTrackingButton = new Lang.Class({
...
             onComplete: Lang.bind(this, function() {
+                Tweener.removeTweens(this);
+                this.rotation_angle_y = 0;
                 Tweener.addTween(this, {
                     opacity: 255,
-                    time: WINDOW_ANIMATION_TIME / 2,
+                    time: WINDOW_ANIMATION_TIME * 0.5,
                     transition: 'linear',
-                    delay: WINDOW_ANIMATION_TIME * 3
+                    delay: WINDOW_ANIMATION_TIME * 1.5
```
smspillaz: Any reason to change these?

js/ui/codingManager.js
```diff
@@ -210,20 +206,72 @@ const WindowTrackingButton = new Lang.Class({
...
                 });
             })
         });
+
+        let button = new St.Bin({ style_class: 'view-source',
+                                  x_fill: true,
+                                  y_fill: false });
+        button.set_child(this._createIcon());
```
smspillaz: Is there any reason why we don't just create the whole button in createIcon?

js/ui/codingManager.js
```diff
@@ -210,20 +206,72 @@ const WindowTrackingButton = new Lang.Class({
...
+
+        Main.layoutManager.addChrome(button);
+        button.set_position(rect.x + rect.width - BUTTON_OFFSET_X,
+                            rect.y + rect.height - BUTTON_OFFSET_Y);
```
smspillaz: This can be probably be shared with the button creation code

js/ui/codingManager.js
```diff
@@ -210,20 +206,72 @@ const WindowTrackingButton = new Lang.Class({
...
+                Main.layoutManager.removeChrome(button);
+                button.destroy();
+            })
+        });
+        Tweener.addTween(button, {
+            opacity: 255,
+            time: WINDOW_ANIMATION_TIME * 2,
+            transition: 'linear',
+            delay: WINDOW_ANIMATION_TIME
+        });
```
smspillaz: Do we also need to add an animation to the other button too? Or just this one?

js/ui/codingManager.js
```diff
@@ -210,20 +206,72 @@ const WindowTrackingButton = new Lang.Class({
...
+    },
+
+    _createIcon: function() {
+        let iconFile = Gio.File.new_for_uri('resource:///org/gnome/shell/theme/rotate.svg');
+        let gicon = new Gio.FileIcon({ file: iconFile });
+        let icon = new St.Icon({ style_class: 'view-source-icon',
+                                 gicon: gicon });
+        return icon;
     },
```
smspillaz: this can be a non-member function since we don't use anything from this class

js/ui/codingManager.js
```diff
@@ -210,20 +214,64 @@ const WindowTrackingButton = new Lang.Class({
...
+        this.rotation_angle_y = 0;
+        Tweener.addTween(this, {
+            rotation_angle_y: direction == Gtk.DirectionType.RIGHT ? 180 : -180,
+            time: WINDOW_ANIMATION_TIME * 4,
+            transition: 'easeOutQuad'
+        });
         Tweener.addTween(this, {
             opacity: 0,
-            time: WINDOW_ANIMATION_TIME / 2,
+            time: WINDOW_ANIMATION_TIME * 2,
```
smspillaz: Is there any reason why we are changing this to be longer?

js/ui/codingManager.js
```diff
@@ -210,20 +214,64 @@ const WindowTrackingButton = new Lang.Class({
...
         });
+
+        let button = new St.Bin({ style_class: 'view-source',
+                                  x_fill: true,
+                                  y_fill: false });
+        button.set_child(_createIcon());
+
+        Main.layoutManager.addChrome(button);
+        button.set_position(rect.x + rect.width - BUTTON_OFFSET_X,
+                            rect.y + rect.height - BUTTON_OFFSET_Y);
```
smspillaz: Did you see my comment earlier about sharing this code with the code that creates the original button?

js/ui/codingManager.js
```diff
@@ -210,20 +214,64 @@ const WindowTrackingButton = new Lang.Class({
...
+        button.rotation_center_y = new Clutter.Vertex({ x: button.width - (rect.width * 0.5),
+                                                        y: button.height * 0.5,
+                                                        z: 0 });
+        button.rotation_angle_y = direction == Gtk.DirectionType.RIGHT ? -180 : 180;
```
smspillaz: It seems to me like we can share a lot of this code as well - the only difference is that the initial angle will be inverted.

js/ui/codingManager.js
smspillaz: This code can probably all be shared.

js/ui/codingManager.js
```diff
@@ -29,239 +46,525 @@ function _isBuilderSpeedwagon(window) {
...
+
+    _init: function(params) {
+        this.parent(params);
+
+        let rect = this.window.get_frame_rect();
+
+        // The button will be auto-added to the manager. Note that in order to
+        // remove this button, you will need to call eject() from outside
+        // this class or use removeChrome from within it.
+        this._button = _createViewSourceButtonInRectCorner(rect);
```
cosimoc: Optional: since `rect` is only used to pass it down here, perhaps you could pass the window directly to that function.

js/ui/codingManager.js
```diff
@@ -29,239 +46,525 @@ function _isBuilderSpeedwagon(window) {
...
+
+        // Do required setup to make this button interactive, by allowing
+        // it to be focused, connecting to the button-press-event signal
+        // and adding chrome to the toplevel layoutManager.
+        this._button.reactive = true;
+        this._button.can_focus = true;
+        this._button.track_hover = true;
+        this._button.connect('button-press-event', Lang.bind(this, function() {
+            this.emit('clicked');
+        }));
```
cosimoc: Optional: could the function that creates the button do this setup? (Or allow a params bag for the button constructor.)

js/ui/codingManager.js
```diff
@@ -29,239 +46,525 @@ function _isBuilderSpeedwagon(window) {
...
+        // this class or use removeChrome from within it.
+        this._button = _createViewSourceButtonInRectCorner(rect);
+
+        // Do required setup to make this button interactive, by allowing
+        // it to be focused, connecting to the button-press-event signal
+        // and adding chrome to the toplevel layoutManager.
+        this._button.reactive = true;
+        this._button.can_focus = true;
+        this._button.track_hover = true;
+        this._button.connect('button-press-event', Lang.bind(this, function() {
```
cosimoc: I believe the button has a `clicked` signal too; this event will also be emitted for e.g. right clicks.

js/ui/codingManager.js
```diff
@@ -29,239 +46,525 @@ function _isBuilderSpeedwagon(window) {
...
+        );
+        this._windowMinimizedId = global.window_manager.connect(
+            'minimize', Lang.bind(this, this._hide)
+        );
+        this._windowUnminimizedId = global.window_manager.connect(
+            'unminimize', Lang.bind(this, this._show)
+        );
+    },
+
+    eject: function() {
```
cosimoc: Nitpick: it's a bit odd to have this function named like this. Perhaps `remove`, or `destroy`?

js/ui/codingManager.js
```diff
@@ -692,41 +984,36 @@ const CodingManager = new Lang.Class({
...
-            actor.opacity = 255;
-            actor.rotation_angle_y = 0;
-            actor.set_cull_back_face(false);
-        }
+    removeBuilderWindow: function(actor) {
+        let window = actor.meta_window;
+        let isSpeedwagonForBuilder = _isBuilderSpeedwagon(window);
+
+        if (!(_isBuilder(window.get_flatpak_id()) ||
+              isSpeedwagonForBuilder))
```
cosimoc: This check is done differently (as an AND of negations) above. Apply De Morgan's law to either :-)

js/ui/codingManager.js
```diff
@@ -652,38 +879,103 @@ const CodingManager = new Lang.Class({
...
+
+    _stopWatchingForBuilderWindowToComeOnline: function() {
+        let tracker = Shell.WindowTracker.get_default();
+        tracker.untrack_coding_app_window();
+        if (this._watchdogId !== 0) {
+            Mainloop.source_remove(this._watchdogId);
+            this._watchdogId = 0;
         }
+
+        this._watchdogId = 0;
```
cosimoc: Remove this

js/ui/codingManager.js
```diff
@@ -29,239 +46,525 @@ function _isBuilderSpeedwagon(window) {
...
+            // point and not at the point we show the speedwagon. This
+            // will ensure that the shell window tracker is still
+            // watching for the builder window to appear.
+            this._stopWatchingForBuilderWindowToComeOnline();
 
-        this._animateToBuilder(session);
-    },
+            // We also only want to hide the speedwagon window at this
+            // point, since the other window has arrived.
+            this.splash.rampOut();
```
cosimoc: Is it guaranteed that you have a splash here?

cosimoc: In other words, does the splash instance stay around if you later receive another builder window? Or do you force a speedwagon every time the button is clicked, regardless of whether Builder is already running?

smspillaz: We force a speedwagon every time the button is clicked.

js/ui/codingManager.js
smspillaz: Maybe - I tried to pick a name that would suggest that we are removing all related state before destruction. `destroy` could also work.

js/ui/codingManager.js
smspillaz: Maybe .... or maybe having a subclass / wrapper class would make sense here? I'll give that a try.

js/ui/codingManager.js
smspillaz: 👍

js/ui/codingManager.js
smspillaz: Yeah - I was wondering if that was the case - I'll give that a try.

smspillaz: Hmm, looks like there is no `clicked` signal on `StBin`

js/ui/codingManager.js
smspillaz: Fixed

js/ui/codingManager.js
smspillaz: Fixed

js/ui/codingManager.js
smspillaz: Err, sorry, no, we force a speedwagon window every time a builder window is about to be associated with the app.

However, on this codepath, `this.splash` is guaranteed to be defined - We are 'admitting' a new builder window actor here and have determined that this window is not a speedwagon. We already created `SpeedwagonSplash` back when the builder window was launched, so here we want to get rid of the `SpeedwagonSplash` instance.

js/ui/codingManager.js
smspillaz: Renamed to destroy

js/ui/codingManager.js
smspillaz: Created a new wrapper function

js/ui/codingManager.js
```diff
@@ -29,239 +46,530 @@ function _isBuilderSpeedwagon(window) {
...
+                                      srcGeometry.height);
+}
+
+function _synchronizeViewSourceButtonToRectCorner(button, rect) {
+    button.set_position(rect.x + rect.width - BUTTON_OFFSET_X,
+                        rect.y + rect.height - BUTTON_OFFSET_Y);
+}
+
+function _createViewSourceButtonInRectCorner(rect) {
+    let button = new St.Bin({
```
cosimoc: I guess I was assuming this was an `St.Button`, not an `St.Bin` -- any reason not to make it such?

PR #247, opened 2017-03-08
coding: Track button to either app or builder window

smspillaz:
Instead of just tracking the app window. This means that we don't
have to wait for a round-trip until the app window gets resized.

https://phabricator.endlessm.com/T15944

### DIFF
--- a/js/ui/codingManager.js
+++ b/js/ui/codingManager.js
@@ -230,11 +230,11 @@ const WindowTrackingButton = new Lang.Class({
     track: function(window) {
         // Untrack the current window and track the newly specified one.
         if (this._positionChangedId) {
-            this._trackingWindow.disconnect(this.positionChangedId);
+            this._trackingWindow.disconnect(this._positionChangedId);
         }
 
         if (this._sizeChangedId) {
-            this._trackingWindow.disconnect(this.sizeChangedId);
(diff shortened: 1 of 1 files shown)


js/ui/codingManager.js
```diff
@@ -208,24 +208,6 @@ const WindowTrackingButton = new Lang.Class({
...
-        this._sizeChangedId = this.window.connect(
-            'size-changed', Lang.bind(this, this._updatePosition)
-        );
-
-        this._windowsRestackedId = Main.overview.connect(
-            'windows-restacked', Lang.bind(this, this._showIfWindowVisible)
-        );
-        this._overviewHidingId = Main.overview.connect(
-            'hiding', Lang.bind(this, this._showIfWindowVisible)
-        );
```
cosimoc: Did you mean to remove these last two?

smspillaz: Oof, no, good catch

smspillaz: Fixed

cosimoc: I think you may have forgotten to push? Looks like these are still missing.

smspillaz: Looks like I did - pushed now.

wjt:
This repo was superseded several years  ago by https://github.com/endlessm/gnome-shell.

PR #252, opened 2017-03-09
CodeView: Report underlying events on CodingSession to interested parties

smspillaz:
For instance, we might be interested in whether the underlying flatpak sources have been downloaded, or if the window has been flipped. One such consumer is coding-game-service, which needs to know about these events in order to move forward with some of the interactions.

https://phabricator.endlessm.com/T13630

### DIFF
--- a/js/ui/codingManager.js
+++ b/js/ui/codingManager.js
@@ -1222,7 +1222,6 @@ const CodeViewManager = new Lang.Class({
 
         // Destroy the session here and remove it from the list
         session.destroy();
-        this._disconnectFlatpakMonitor();
 
         let idx = this._sessions.indexOf(session);
         if (idx != -1) {
--- a/js/ui/codingManager.js
+++ b/js/ui/codingManager.js
@@ -12,7 +12,6 @@ const Mainloop = imports.mainloop;
 const Meta = imports.gi.Meta;
 const Shell = imports.gi.Shell;
 const St = imports.gi.St;
-const IconGridLayout = imports.ui.iconGridLayout;
 
 const AppActivation = imports.ui.appActivation;
 const Main = imports.ui.main;
@@ -1076,6 +1075,7 @@ const CodingOSIntegration = new Lang.Class({
                 }
             }));
 
+            session.connect('closed', Lang.bind(this, function() {
                 if (listeningFor.has('codeview-closed')) {
                     listeningFor.delete('codeview-closed');
                     controller.event_occurred('codeview-closed');
--- a/js/ui/codingManager.js
+++ b/js/ui/codingManager.js
@@ -467,6 +467,14 @@ const CodingSession = new Lang.Class({
                                                     GObject.ParamFlags.CONSTRUCT_ONLY,
                                                     CodingInstallationMonitor.$gtype)
     },
+    Signals: {
+        'opened': {},
+        'ready': {},
+        'closed': {},
+        'flipped': {
+            param_types: [ GObject.TYPE_INT ]
+        }
+    },
 
     _init: function(params) {
         this.parent(params);
@@ -534,6 +542,7 @@ const CodingSession = new Lang.Class({
             // We also only want to hide the speedwagon window at this
             // point, since the other window has arrived.
             this.splash.rampOut();
+            this.emit('ready');
         }
 
         // Now, if we're not already on the builder window, we want to start
@@ -560,6 +569,7 @@ const CodingSession = new Lang.Class({
                                  this.builder,
                                  Gtk.DirectionType.LEFT);
             this._state = STATE_BUILDER;
+            this.emit('flipped', this._state);
         }
 
         return true;
@@ -585,9 +595,11 @@ const CodingSession = new Lang.Class({
         this.button.builder_window = null;
         this.builder = null;
         this._state = STATE_APP;
+        this.emit('flipped', this._state);
 
         this.app.meta_window.activate(global.get_current_time());
         this.app.show();
+        this.emit('closed');
     },
 
     // Eject out of this session and remove all pairings.
@@ -642,6 +654,7 @@ const CodingSession = new Lang.Class({
 
             this.builder.meta_window.activate(global.get_current_time());
             this.builder.show();
+            this.emit('closed');
 
             // Note that we do not set this._state to STATE_APP here. Any
             // further usage of this session is undefined and it should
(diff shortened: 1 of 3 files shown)


js/ui/codingManager.js
```diff
@@ -1076,6 +1075,7 @@ const CodingOSIntegration = new Lang.Class({
                 }
             }));
 
+            session.connect('closed', Lang.bind(this, function() {
```
smspillaz: Oof, that's awkward. I must have missed that in a rebase or something O.o

js/ui/codingManager.js
```diff
@@ -359,6 +361,84 @@ const WindowTrackingButton = new Lang.Class({
     }
 });
 
+const CodingInstallationMonitor = new Lang.Class({
+    Name: 'CodingInstallationMonitor',
+    Extends: GObject.Object,
+    Properties: {
+        'flatpak-name': GObject.ParamSpec.string('flatpak-name',
```
cosimoc: This should be called `flatpak-id` or something like that -- `name` usually implies something user-visible.

js/ui/codingManager.js
```diff
@@ -359,6 +361,84 @@ const WindowTrackingButton = new Lang.Class({
...
+
+        let userInstallation = Flatpak.Installation.new_user(null);
+
+        let apps = userInstallation.list_installed_refs_by_kind(Flatpak.RefKind.APP, null);
+        for (let i = 0; i < apps.length; i++) {
+            if (apps[i].name === params.flatpak_name) {
+                this._commit = apps[i].commit;
+                break;
+            }
+        }
```
cosimoc: Should we do something here when the app is not found?

js/ui/codingManager.js
```diff
@@ -359,6 +361,84 @@ const WindowTrackingButton = new Lang.Class({
...
+                // we are monitoring a file .changed which
+                // gets deleted and created on every change independent
+                // of install or uninstall, therefore only track the done hint
+                if (type !== Gio.FileMonitorEvent.CHANGES_DONE_HINT)
+                    return;
+
+                let app;
+                try {
+                    app = userInstallation.get_current_installed_app(this.flatpak_name, null);
+                } catch(e) {
```
cosimoc: Space before paren

js/ui/codingManager.js
```diff
@@ -359,6 +361,84 @@ const WindowTrackingButton = new Lang.Class({
...
+                    // application not installed or just got deleted
+                    return;
+                }
+
+                if (app.commit === this._commit)
+                    return;
+
+                this._commit = app.commit;
+                this._changed = true;
```
cosimoc: You should notify the `installed` property here.

js/ui/codingManager.js
```diff
@@ -359,6 +361,84 @@ const WindowTrackingButton = new Lang.Class({
...
+
+    disconnect: function() {
+        if (this._flatpakMonitorId) {
+            this._flatpakMonitor.disconnect(this._flatpakMonitorId);
+            this._flatpakMonitorId = 0;
+            this._flatpakMonitor = null;
+        }
+    },
+
+    get installed() {
```
cosimoc: It strikes me that this is not a good name for this property.

js/ui/codingManager.js
```diff
@@ -380,7 +460,24 @@ const CodingSession = new Lang.Class({
...
+    },
+    Signals: {
+        'ready': {},
+        'closed': {},
+        'flipped': {
+            param_types: [ GObject.TYPE_INT ]
+        }
+    },
+
+    get state() {
```
cosimoc: Do you want this to be a property instead? (and remove the `flipped` signal, since you could notify on it)

js/ui/codingManager.js
```diff
@@ -942,10 +1213,16 @@ const CodingManager = new Lang.Class({
...
-            button: new WindowTrackingButton({ window: actor.meta_window })
-        }));
+            button: new WindowTrackingButton({ window: actor.meta_window }),
+            install_monitor: new CodingInstallationMonitor({
+                flatpak_name: window.get_flatpak_id()
+            })
+        });
+        this._sessions.push(session);
+
+        this.emit('session-created', session);
```
cosimoc: Instead of using this signal here, you can just add a method to the `CodingOSIntegration` object and call it; no other client either listens to this or uses `CodingOSIntegration`

js/ui/codingManager.js
```diff
@@ -1021,5 +1298,9 @@ const CodingManager = new Lang.Class({
         }
 
         return null;
-    }
+    },
+
+    _isCodingApp: function(flatpakID) {
```
cosimoc: I don't think this is required here.

js/ui/codingManager.js
```diff
@@ -927,11 +1035,174 @@ const CodingSession = new Lang.Class({
...
+    get sessions() {
+        return this._sessions;
+    },
 
-    _init: function() {
+    _init: function(params) {
+        this.parent(params);
         this._sessions = [];
+
+        this._codingApps = ['com.endlessm.Helloworld', 'org.gnome.Weather.Application'];
```
cosimoc: I think this can be removed.

js/ui/codingManager.js
```diff
@@ -927,11 +1035,174 @@ const CodingSession = new Lang.Class({
...
+
+const CodingOSIntegration = new Lang.Class({
+    Name: 'CodingOSIntegration',
+
+    _init: function(manager) {
+        let listeningFor = Set();
+
+        manager.connect('session-created', Lang.bind(this, function(instance, session) {
+            // Check to see if the session was created on org.gnome.Weather.Application
+            if (session.app.meta_window.get_flatpak_id() !== OS_INTEGRATION_APP)
```
cosimoc: I don't understand this check here; my understanding is that if we got here, we are already on a session for one of the whitelisted apps.

js/ui/codingManager.js
```diff
@@ -927,11 +1035,174 @@ const CodingSession = new Lang.Class({
     }
 });
 
-const CodingManager = new Lang.Class({
-    Name: 'CodingManager',
+// Given a FlatpakInstalledRef app, look up its exports directory
+// and find any relevant .desktop files and add them to the desktop
+// if they are not already added.
+function addFlatpakAppToDesktop(app) {
```
cosimoc: I am not sure this function is needed; you should just be able to call `appendIcon` with the app ID. Basically use the same exact behavior the app store has when an app is installed, which ends up here: https://github.com/endlessm/eos-desktop/blob/master/js/ui/shellDBus.js#L493

cosimoc (changes requested):

wjt:
This repo was superseded several years  ago by https://github.com/endlessm/gnome-shell.

PR #254, opened 2017-03-13
updaterManager: Ignore errors from running on live booted systems

pwithnall:
eos-updater will not update a live-booted system, so do not show errors
from it saying the system is live booted.

Signed-off-by: Philip Withnall <withnall@endlessm.com>

https://phabricator.endlessm.com/T15977


cosimoc:
@wjt Update bubbles were generally broken for a while -- see https://phabricator.endlessm.com/T16001
But the updater also now reports errors more consistently than it used to. Either way this looks good to me as well.

PR #255, opened 2017-03-14
coding: Reduce timeout for wobbly_window demo to 2500

smspillaz:
https://phabricator.endlessm.com/T16114


erikos:
Sounds good, I actually had the same intuition when testing today...

PR #257, opened 2017-03-22
messageTray: Hide overview on notification activation

krnowak:
Pressing the notification when the overview was in front of all the
non-minimized windows of other applications didn't result in the
application window being shown. Hide the overview explicitly then.

https://phabricator.endlessm.com/T16208

This looks a bit like a blunt way to hide the overview, but I noticed that there was a prior art in fdo notifications handling, so I followed suit.


krnowak:
Activate the app then hide the overview. You can have a look here what does fdo backend do - https://github.com/kinvolk/eos-desktop/blob/74bd2f63fb70b6788e626a02bfe6732d99555328/js/ui/notificationDaemon.js#L688-L694

PR #258, opened 2017-04-04
Add runtime dependency on mutter, in preparation for the shell rebase

mariospr:
GNOME Shell 3.22 will depend on the mutter package, which will contain
the mutter-restart-helper tool that we use to ship in mutter-common with
a downstream patch we no longer have in mutter 3.22. For this reason,
make our current shell depend explicitly on mutter already now, without
waiting for GNOME Shell 3.22, so that the helper gets installed too.

https://phabricator.endlessm.com/T16276


manuq:
@mariospr quick question: I thought the eos-desktop was a whole fork from GNOME Shell. but from this pull request should I conclude that eos-desktop is being rebased to contain latest changes from gnome-shell? Or is this only about the mutter helper being upstreamed?

mariospr:
You're right: eos-shell is a fork of GNOME Shell 3.8 and I'm right now working precisely on the "Big Rebase" to the 3.22 codebase. This is why I've been rebasing several components that we need (gdm, mutter) lately, so that I can start porting our downstream changes on top of the new codebase, because there are so many things to port that, clearly, calling it "a rebase" is kind of an understatement :)

That said, this PR is not meant to be merged yet, I need to iron out a few details with mutter first, but I hope will be done later today or tomorrow.

manuq:
@mariospr thanks for the explanation.

mariospr:
As per some internal discussion we might not be depending on the mutter package after all, so I'm closing this PR for now, while I do some more experiments.

PR #260, opened 2017-04-12
GrandCentral: slide down app window from the top center of the screen

erikos:


js/ui/windowManager.js
```diff
@@ -834,27 +834,44 @@ const WindowManager = new Lang.Class({
...
-                           time: WINDOW_ANIMATION_TIME,
-                           transition: "easeOutQuad",
-                           onComplete: onComplete,
-                           onCompleteScope: this,
-                           onCompleteParams: [shellwm, actor],
-                           onOverwrite: onOverwrite,
-                           onOverwriteScope: this,
-                           onOverwriteParams: [shellwm, actor]
-                         });
+        if (actor.x == centered) {
```
smspillaz: Checking to see if the actor is centered on the monitor like this seems awfully fragile - can we check some other flag on the window?

js/ui/windowManager.js
```diff
@@ -1059,29 +1076,53 @@ const WindowManager = new Lang.Class({
...
-        if (origX == monitor.x) {
-            // the side bar will appear from the left side
-            actor.set_position(monitor.x - actor.width, actor.y);
-        } else {
-            // ... from the right side
-            actor.set_position(monitor.x + monitor.width, actor.y);
+        let centered = (monitor.width - actor.width) * 0.5;
+
+        if (actor.x == centered) {
+            // the side bar will appear from the top center
```
smspillaz: It isn't exactly a side bar if it appears from the top center ;-)

js/ui/windowManager.js
erikos: Yeah, I have added now a helper function to check the wm_class if it is the GC window and make the decision based on that.

js/ui/windowManager.js
erikos: Fixed :)

js/ui/windowManager.js
```diff
@@ -1059,29 +1074,51 @@ const WindowManager = new Lang.Class({
...
+                actor.set_position(monitor.x - actor.width, actor.y);
+            } else {
+                // ... from the right side
+                actor.set_position(monitor.x + monitor.width, actor.y);
+            }
 
-        actor.opacity = 255;
-        actor.show();
+            actor.opacity = 255;
+            actor.show();
```
smspillaz: You can probably call this outside the if conditions, since it is repeated in both cases.

erikos: I want this to happen after I set the position that is why I repeated it in both code paths.

smspillaz: They can probably be called on line 1085 though - it shouldn't be a problem if they are called after `Tweener.addTween` since that is asynchronous anyway.

smspillaz:
Code looks good to me, just going to give it a quick test now

PR #262, opened 2017-04-13
Adapt eos-shell 3.8 to build and run on top of mutter 3.22 (debian)

mariospr:
Some changes are required in the debian packaging to build eos-shell on top of the newer mutter, so we borrow some bits from debian unstable's gnome-shell_3.22.3 package which, besides fixing the packaging issues now that clutter and cogl are shipped inside of mutter, it should also make things easier when rebasing our old shell on top of GNOME shell 3.22.

https://phabricator.endlessm.com/T16276


debian/rules
```diff
@@ -1,19 +1,42 @@
...
+override_dh_girepository:
+	dh_girepository \
+		-l src:src/gvc \
+		-p /usr/lib/$(DEB_HOST_MULTIARCH)/mutter \
+		/usr/lib/eos-shell
 
-# List any files which are not installed
-common-binary-post-install-arch:: list-missing
+override_dh_strip:
+	dh_strip --dbgsym-migration='eos-shell-dbg (<< 3.20.2-2~)'
```
cosimoc: We're not bumping the version of `eos-shell` to 3.20 as part of this though, right?

mariospr: Good catch, I'll remove that.

PR #264, opened 2017-04-20
DiscoveryFeed: add tale to desktop

erikos:
https://phabricator.endlessm.com/T16635


js/js-resources.gresource.xml
```diff
@@ -112,12 +112,13 @@
     <file>ui/components/appStore.js</file>
     <file>ui/components/autorunManager.js</file>
     <file>ui/components/automountManager.js</file>
-    <file>ui/components/keyring.js</file>
+    <file>ui/components/codingManager.js</file>
     <file>ui/components/codingGameService.js</file>
+    <file>ui/components/discoveryFeed.js</file>
+    <file>ui/components/keyring.js</file>
```
smspillaz: Can probably revert the change here which moves `keyring.js` around

js/ui/viewSelector.js
```diff
@@ -53,16 +53,18 @@ const ViewsDisplayLayout = new Lang.Class({
     Name: 'ViewsDisplayLayout',
     Extends: Clutter.BinLayout,
 
-    _init: function(entry, allViewActor, searchResultsActor) {
+    _init: function(entry, gc, allViewActor, searchResultsActor) {
         this.parent();
 
         this._entry = entry;
+        this._gc = gc;
```
smspillaz: Probably rename this to `discoveryFeed` given that we are not using the old codename anymore.

js/ui/viewSelector.js
```diff
@@ -53,16 +53,18 @@ const ViewsDisplayLayout = new Lang.Class({
...
         this._allViewActor = allViewActor;
         this._searchResultsActor = searchResultsActor;
 
         this._entry.connect('style-changed', Lang.bind(this, this._onStyleChanged));
         this._allViewActor.connect('style-changed', Lang.bind(this, this._onStyleChanged));
 
+        this._heightAboveGC = 0;
```
smspillaz: It seems like this is always zero. Can it be made a constant instead of an instance variable?

js/ui/viewSelector.js
```diff
@@ -284,7 +303,19 @@ const ViewsDisplay = new Lang.Class({
...
+        let icon = new St.Icon({ icon_name: 'edit-find-symbolic',
+                                style_class: 'discovery-feed-icon',
+                                icon_size: 16,
+                                track_hover: true });
+        this.gc = new St.Button({ style_class: 'discovery-feed',
+                                  x_align: Clutter.ActorAlign.CENTER,
+                                  y_align: Clutter.ActorAlign.CENTER });
+        this.gc.set_child(icon);
+        this.gc.connect('clicked', Lang.bind(this, function() {
+          Main.discoveryFeed.show(0);
```
smspillaz: Should we be calling this with an actual timestamp instead of zero?

js/ui/viewSelector.js
```diff
@@ -284,7 +303,19 @@ const ViewsDisplay = new Lang.Class({
...
+        }));
+
+        this.actor = new ViewsDisplayContainer(this.entry, this.gc, this._allView, this._searchResults);
```
smspillaz: What is the difference between `this.gc` which is passed to `ViewsDisplayContainer` and `gc` passed to `ViewsDisplayLayout` below?

js/ui/viewSelector.js
```diff
@@ -132,6 +139,16 @@ const ViewsDisplayLayout = new Lang.Class({
...
 
+        let gcHeight = this._gc.get_preferred_height(availWidth)[1];
+
+        let gcBox = box.copy();
+        let x1 = (availWidth - this._gc.get_width()) * 0.5;
+        gcBox.y1 = this._heightAboveGC;
+        gcBox.y2 = gcBox.y1 + gcHeight;
+        gcBox.x1 = x1;
+        gcBox.x2 = x1 + this._gc.get_width();
+        this._gc.allocate(gcBox, flags);
```
smspillaz: While this is consistent with the existing code, my preference here would be to create a subclass of `St.Button` for the discovery feed handle and encapsulate as much of its code as possible inside of instance methods. We would only need to pass `availWidth` and `this._heightAboveGC` to it.

js/ui/viewSelector.js
smspillaz: See comment above about encapsulating a lot of this logic into its own subclass of `St.Button`

smspillaz:
Tested, everything seems to work fine with the gsetting and using the corresponding branch on e-d-f

js/ui/components/discoveryFeed.js
```diff
@@ -0,0 +1,49 @@
...
+const Lang = imports.lang;
+
+const Main = imports.ui.main;
+const SideComponent = imports.ui.sideComponent;
+
+const DISCOVERY_FEED_NAME = 'com.endlessm.DiscoveryFeed';
+const DISCOVERY_FEED_PATH = '/com/endlessm/DiscoveryFeed';
+
+const DiscoveryFeedIface = '<node> \
+<interface name="' + DISCOVERY_FEED_NAME + '"> \
```
smspillaz: For consistency here, I'd add a `const DISCOVERY_FEED_IFACE` that explicitly specifies the interface name. Some of the trouble I was debugging last night was because of confusion between the dbus name and the interface

js/ui/viewSelector.js
```diff
@@ -416,6 +476,8 @@ const ViewsClone = new Lang.Class({
                                                           enabled: false });
         iconGridClone.add_effect(this._saturation);
 
+        if (discoveryFeed !== null)
+            this.add_child(discoveryFeed);
```
smspillaz: Instead of checking whether discoveryFeed is null here, you can probably just add the child in the parent if condition

erikos: I actually tried that, but it has to happen after the constructing of the ViewsDisplayLayout.

PR #267, opened 2017-06-12
DiscoveryFeed: make the window semi-transparent

erikos:
This should have the same values as the search
result widget.

### DIFF
--- a/js/ui/windowManager.js
+++ b/js/ui/windowManager.js
@@ -36,7 +36,8 @@ const DIM_BRIGHTNESS = -0.3;
 const DIM_TIME = 0.500;
 const UNDIM_TIME = 0.250;
 const SKYPE_WINDOW_CLOSE_TIMEOUT_MS = 1000;
-
+const DISCOVERY_FEED_OPACITY = 200;
+const FULL_OPACITY = 255;
 const ONE_SECOND = 1000; // in ms
 
 const DisplayChangeDialog = new Lang.Class({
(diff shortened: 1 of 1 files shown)


smspillaz:
Hmm, my preference here would be to set an actual transparency hint on the GtkWindow itself. I believe that we're doing that within the app itself, though it doesn't seem to be working :/

cosimoc (changes requested):
I agree, this should not be done at the window manager level but within the app.
I believe the old app store did that, so we may be able to take inspiration from that app for what we need to do here.

erikos:
Right, setting the opacity on the window itself in the app was the first thing I tried and we actually had already a setting there but it is not picked up. The window is composited like noted in the doc to be needed. I did try to set the opacity for a gtkwindow as well in a test case but it does not seem to be picked up.

As we do set opacity to full when done with window mapping in the windowManager.js I special cased those to not be set in the DiscoveryFeed case to make sure we do not override opacity but same result, that setting is not propagated it looks like.

Looked at the apps-store code but could not find a special casing there yet regarding opacity.

cosimoc:
@erikos Not sure if setting the opacity on the window itself is the way to go, but I believe you can tweak the background color of the main GtkWindow using CSS.

cosimoc:
(And just to confirm, the opacity that you would set from GTK is never going to be reflected in the opacity values you see in `windowManager.js` -- they're unrelated.)

erikos:
Thanks Cosimo for your comments - new approach inspired by eos-app-store https://github.com/endlessm/eos-discovery-feed/pull/31

PR #269, opened 2017-06-19
DiscoveryFeed: add hover state for tile

erikos:
This also makes the tile itself a bit bigger.


erikos:
Moved to new shell: https://github.com/endlessm/gnome-shell/pull/54

